PR #83, opened 2016-03-25
More doc changes

sukidog:


c1728p9:
@sukidog can you please sign https://developer.mbed.org/contributor_agreement/ ? I was not able to locate you on mbed.org, so you might need to make an account.

c1728p9:
Thanks for the doc updates.  They look good.

PR #86, opened 2016-03-29
Fix --notestendpt handling

sukidog:
The test logic was requiring a target executable even when --notestendpt was specified. The target binary is used only by the endpoint testing.

### DIFF
--- a/test/daplink_firmware.py
+++ b/test/daplink_firmware.py
@@ -80,7 +80,19 @@ class ProjectFirmwareBundle(firmware.FirmwareBundle):
     def __init__(self, directory, tool):
         tool_dir = os.path.abspath(directory + os.sep + 'projectfiles' +
                                    os.sep + tool)
+        errmsg = "Error: DAPLink projects are missing. Did " \
+                   "you forget to generate them by running " \
+                   "'progen generate -t %s'?\nExpecting them at %s" % (tool, tool_dir)
+
+        if not os.path.exists(tool_dir): 
+            print (errmsg)
+            exit(-1) 
+
         project_dir_list = os.listdir(tool_dir)
+        if not project_dir_list:
+            print (errmsg)
+            exit(-1) 
+        
         firmware_list = []
         for name in project_dir_list:
             build_dir = tool_dir + os.sep + name + os.sep + 'build'
--- a/test/run_test.py
+++ b/test/run_test.py
@@ -431,7 +431,7 @@ class TestManager(object):
 
             # Get target
             target = None
-            target_required = not self._test_ep and not self._test_daplink
+            target_required = self._test_ep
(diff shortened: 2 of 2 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

test/daplink_firmware.py
```diff
@@ -80,7 +80,19 @@ class ProjectFirmwareBundle(firmware.FirmwareBundle):
...
+                   "'progen generate -t %s'?\nExpecting them at %s" % (tool, tool_dir)
+
+        if not os.path.exists(tool_dir): 
+            print (errmsg)
+            exit(-1) 
+
         project_dir_list = os.listdir(tool_dir)
+        if not project_dir_list:
+            print (errmsg)
+            exit(-1)
```
c1728p9: It might be better to raise and exception here or add a 'vaild' field rather than calling exit.  That way the script using this is able to recover from an error rather than exiting

test/run_test.py
```diff
@@ -573,25 +573,32 @@ def main():
...
-    else:
-        target_dir = daplink_dir + os.sep + 'tmp'
-        build_target_bundle(target_dir, args.user, args.password, test_info)
+    
+        if args.targetdir is not None:
+            target_dir = args.targetdir
+        else:
+            target_dir = daplink_dir + os.sep + 'tmp'
+            build_target_bundle(target_dir, args.user, args.password, test_info)
+
```
c1728p9: Nitpick - trailing whitespace on line 579, 593 and 599

test/daplink_firmware.py
sukidog: I thought about that but then noticed there are a number of exit(-1) calls in run_test.py. I was trying to be consistent. Are you saying to throw an exception so the calling *function* can continue if it wishes to?

test/run_test.py
sukidog: Good nitpick. I found four other lines with trailing whitespace. Cleanup pushed

test/daplink_firmware.py
c1728p9: By script I meant run_test.py.  Currently everything which uses exit() resides in run_test.py and isn't meant to be used outside that script.  This is more of a nitpick so don't feel like it needs to change.  If the need arises in the future to make this case recoverable then it can be changed later.

c1728p9:
@mbed-bot: TEST

test/daplink_firmware.py
sukidog: Sounds good. I will leave it as is. Doesn't seem like a recoverable error condition. Like you said, we can change that as needed

John

c1728p9:
Thanks for the improvements!

PR #87, opened 2016-03-29
uVision pre/post script failure did not fail build

sukidog:
The daplink post/pre scripts invoked by the uVision builds lacked proper error checking and reporting. The projects, in turn, didn't care of the scripts failed. Both of these deficiencies are fixed by this PR. Also improved script output--removed a lot of clutter.


sukidog:
This got a bit messy. Creating a new PR

Issue #47, opened 2016-01-29
venv hides tools


sg-:
Just putting this here because I think we had an idea how to fix it but I don't recall the solution.

brianesquilona:
NA to current env.

PR #88, opened 2016-04-01
uVision pre/post script failure do not fail build

sukidog:
The daplink post/pre scripts invoked by the uVision builds lacked proper error checking and reporting. The projects, in turn, didn't care of the scripts failed. Both of these deficiencies are fixed by this PR. Also improved script output--removed a lot of clutter.

Also added script for launching uVision in the proper environment from the cmdline. Updated documentation.

### DIFF
--- a/records/tools/uvision.uvproj.tmpl
+++ b/records/tools/uvision.uvproj.tmpl
@@ -62,8 +62,6 @@
             <UserProg2Name></UserProg2Name>
             <UserProg1Dos16Mode>0</UserProg1Dos16Mode>
             <UserProg2Dos16Mode>0</UserProg2Dos16Mode>
-            <nStopU1X>0</nStopU1X>
-            <nStopU2X>0</nStopU2X>
           </BeforeCompile>
           <BeforeMake>
             <RunUserProg1>1</RunUserProg1>
@@ -72,6 +70,7 @@
             <UserProg2Name></UserProg2Name>
             <UserProg1Dos16Mode>0</UserProg1Dos16Mode>
             <UserProg2Dos16Mode>0</UserProg2Dos16Mode>
+            <nStopB1X>2</nStopB1X>
           </BeforeMake>
           <AfterMake>
             <RunUserProg1>1</RunUserProg1>
@@ -80,6 +79,7 @@
             <UserProg2Name></UserProg2Name>
             <UserProg1Dos16Mode>0</UserProg1Dos16Mode>
             <UserProg2Dos16Mode>0</UserProg2Dos16Mode>
+            <nStopA1X>2</nStopA1X>
           </AfterMake>
           <SelectedForBatchBuild>0</SelectedForBatchBuild>
           <SVCSIdString></SVCSIdString>
--- a/tools/build_release_uvision.bat
+++ b/tools/build_release_uvision.bat
@@ -47,7 +47,7 @@ pip freeze                                      > uvision_release\build_requirem
 
 @REM build but continue if there are errors
 progen generate -t uvision -b
-@SET level=%errorlevel%
+@set level=%errorlevel%
 python tools/copy_release_files.py
 @if %errorlevel% neq 0 exit /B %errorlevel%
 
--- a/tools/build_release_uvision.bat
+++ b/tools/build_release_uvision.bat
@@ -16,11 +16,8 @@
 :: limitations under the License.
 ::
 
-@if exist .gitignore goto label0
-	@echo Error: working directory must be root of DAPLink repository
-	@exit /B 1
-
-:label0
+@rem Script assumes working directory is workspace root. Force it.
+cd %~dp0..\
 
 rmdir /q /s uvision_release
 @if %errorlevel% neq 0 exit /B %errorlevel%
--- a/tools/launch_uv4.bat
+++ b/tools/launch_uv4.bat
@@ -15,6 +15,12 @@
 :: See the License for the specific language governing permissions and
 :: limitations under the License.
 ::
+:: Launches uVision with the python environment needed to build DAPLink
+::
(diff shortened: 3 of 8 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

tools/build_release_uvision.bat
```diff
@@ -15,22 +15,40 @@
...
-REM use custom requirements if specified
+@REM use custom requirements if specified
 if not [%1]==[] pip install -r %1
 pip freeze                                      > uvision_release\build_requirements.txt
-REM build but continue if there are errors
+
+@REM build but continue if there are errors
 progen generate -t uvision -b
-SET LEVEL=%ERRORLEVEL%
+@SET level=%errorlevel%
```
c1728p9: Was caps causing problems, or is this just a cosmetic change?

sukidog: It was not causing a problem. But the scripts were inconsistent and I find
that distracting, and I tried to bring consistency at least to the scripts
I was touching. Uppercase vs lowercase--I don't really have a preference. I
arbitrarily picked one and I can easily switch to the other and make it
consistent if you like

c1728p9: Just wanted to make sure there wasn't a subtle bug that this fixed.  (I don't do much batch scripting so I could have easily created a bug).  Case doesn't matter to me either.  Keeping everything consistent as you did is preferred.

c1728p9:
Can you rebase this patch on top of master?  It has conflicts with the latest merge into master.

tools/build_release_uvision.bat
```diff
@@ -15,22 +15,40 @@
 :: See the License for the specific language governing permissions and
 :: limitations under the License.
 ::
+
```
c1728p9: If you change to the root directory of the project at the start of the script then you can run this from anywhere.  To do this "cd %~dp0..\" would be needed here.

tools/launch_uv4.bat
```diff
@@ -0,0 +1,58 @@
...
+:: See the License for the specific language governing permissions and
+:: limitations under the License.
+::
+:: Launches uVision with the python environment needed to build DAPLink
+::
+:: git and python are expected to be in PATH. Project will fail to build otherwise
+:: 
+
+setlocal
+
```
c1728p9: Same as above, if directory is changed here then you can run this script from anywhere

tools/launch_uv4.bat
```diff
@@ -0,0 +1,58 @@
...
+call env\Scripts\activate
+
+@echo Doing pip install
+@REM use project requirements if not specified
+if [%1]==[] pip install -r requirements.txt
+@REM use custom requirements if specified
+if not [%1]==[] pip install -r %1
+
+:launch
+%uv4exe%
```
c1728p9: If this is replaced with "start %uv4exe%" then it will start in a new process with the virtualenv still active.  This in addition to letting the script run anywhere would allow a user to just double click launch_uv4.bat to start it.

tools/pre_build_script.bat
```diff
@@ -16,12 +16,16 @@
 :: limitations under the License.
 ::
 
-set TOOLS=..\..\..\tools
-set DIR=..\..\..\source\daplink
-REM  Create default version_git.h.  This is so the project 
-REM  builds if python is not installed
+@set TOOLS=..\..\..\tools
```
c1728p9: I forgot you could supress individual messages with @.  This is a great idea!  It makes the build output much cleaner

tools/launch_uv4.bat
```diff
@@ -0,0 +1,58 @@
+::
```
c1728p9: Nitpick - might want to call this launch_uvision.bat rather than launch_uv4.  On my machine this starts up uvision5

records/tools/uvision.uvproj.tmpl
```diff
@@ -72,6 +70,7 @@
             <UserProg2Name></UserProg2Name>
             <UserProg1Dos16Mode>0</UserProg1Dos16Mode>
             <UserProg2Dos16Mode>0</UserProg2Dos16Mode>
+            <nStopB1X>2</nStopB1X>
```
c1728p9: I don't think uvision 4 respects the stop on error setting.  I can confirm that uvision 5 stops on an error correctly.  I think uv4 is just missing this feature.  I'm not sure if much can be done about this, just figured I would mention it.

c1728p9:
Thanks for the patch!  There are a lot of good changes here.

Also it looks like the launch_uv4.bat script will take care of issue #47.

tools/launch_uv4.bat
sukidog: Good idea. Will implement

tools/launch_uv4.bat
sukidog: The crazy thing is the MDKv5 has uv4.exe in a UV4 directory! But good idea. I'll rename the script

records/tools/uvision.uvproj.tmpl
sukidog: Bummer. Oh well, hopefully most people use v5 and benefit from this. I assume the added elements will just be ignored by v4. Can you confirm? (I don't have it installed)

c1728p9: Tested with uv4 and it doesn't have any problems with these new fields.

tools/launch_uv4.bat
sukidog: BTW, the script does not require that you be in the virtual env when launching it. However, using 'start' to launch the executable does provide the benefit that you can continue using the shell (venv or not)

PR #89, opened 2016-04-07
Document TARGET_NAME_TO_BOARD_ID

sukidog:
I scratched my head quite a bit when I first saw this map. Only after learning the code did it make sense.

### DIFF
--- a/test/info.py
+++ b/test/info.py
@@ -62,6 +62,14 @@ FIRMWARE_NAME_TO_BOARD_ID = {
     'lpc11u35_archlink_if': 0x9013,
 }
 
+# Maps a target application filename (sans extension) to the ID of the board 
+# it is built to run on. This mapping is also used to display a board name 
+# in the test output. Thus, name the target applications after the  board to
+# avoid confusing results.
+#
+# One-to-many is fine. Many-to-one is fine as long as only one (or none) target 
+# application is found on disk per connected board ID
+#
 TARGET_NAME_TO_BOARD_ID = {
     'FRDM-K22F': 0x0231,
     'NXP-LPC800-MAX': 0x1050,
--- a/test/info.py
+++ b/test/info.py
@@ -62,13 +62,17 @@ FIRMWARE_NAME_TO_BOARD_ID = {
     'lpc11u35_archlink_if': 0x9013,
 }
 
-# Maps a target application filename (sans extension) to the ID of the board 
-# it is built to run on. This mapping is also used to display a board name 
-# in the test output. Thus, name the target applications after the  board to
-# avoid confusing results.
+# Maps the target test application to the ID of the board it is built to run 
+# on.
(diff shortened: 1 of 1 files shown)


c1728p9:
@mbed-bot: test

c1728p9:
@mbed-bot: TEST

test/info.py
```diff
@@ -62,6 +62,14 @@
     'lpc11u35_archlink_if': 0x9013,
 }
 
+# Maps a target application filename (sans extension) to the ID of the board
```
c1728p9: Is 'sans' a typo?

test/info.py
```diff
@@ -62,6 +62,14 @@
...
 }
 
+# Maps a target application filename (sans extension) to the ID of the board 
+# it is built to run on. This mapping is also used to display a board name 
+# in the test output. Thus, name the target applications after the  board to
+# avoid confusing results.
+#
+# One-to-many is fine. Many-to-one is fine as long as only one (or none) target 
+# application is found on disk per connected board ID
+#
```
c1728p9: Currently all the entries in TARGET_NAME_TO_BOARD_ID correspond to the boards on https://developer.mbed.org/platforms/<board name here>.  The compile API used for testing is based on the URL of the platform page for the board.  For example, the 'NXP-LPC800-MAX' page can be found here https://developer.mbed.org/platforms/NXP-LPC800-MAX/.

test/info.py
sukidog: No. sans=without. That map has as its key the file name of the target
application minus the extension.

test/info.py
sukidog: I have a TARGET_NAME_TO_BOARD_ID entry in our local git repository for our
KL28Z boards, and those boards are not mbed platforms. The target
application binary is found using --targetdir.

sukidog: I guess what you're saying, though, is that for the mbed cases, the file name cannot be arbitrary--it must match what's on the mbed site. If so, then, I would indeed include that in the comment. I can add that or you can.

c1728p9: Yeah, it need to match for the compile API to work.  If the image was created with by some other means then the name doesn't matter

c1728p9: If it isn't easy to update the comment then I'll take care of it in another PR.

c1728p9:
This patch looks good to me.  Shall I merge it?

test/info.py
sukidog: I'll update the comment now...

sukidog: OK. I pushed an updated comment. I took out the key/map relationship comment since it was written only with --targetdir in mind and I'm not sure what is allowed in the Compile API case. Not a big deal. I think 1:1 is what will mostly be used.

c1728p9:
Thanks for the doc updates

PR #92, opened 2016-04-19
Micro:Bit DAPLink changes

alexk2k12:
## Micro:Bit DAPLink changes

Quick overview: the BBC Micro:Bit (https://www.microbit.co.uk/) has DAPLink running on a KL26Z to program an NRF51822. A few extra features are required:
 1. The ability to preserve some pages in flash, across rewrites. For example, to preserve BLE state information and sensor calibration data - saves having to set these every time the device is flashed.
 2. The codebase for the NRF51822 has a filesystem implementation, to save user data to flash. A method is needed to access this information via USB on a computer.

The latter is achieved by presenting the contents of flash as a javascript array. An HTML webpage is also made available, which imports this javascript, parses the filesystem and presents the files.

### Changes:
 - Created a new directory: source/board/microbit for files specific to the microbit.
 - The definition of flash_intf in target_flash.c checks a macro: TARGET_FLASH_ERASE_CHIP_OVERRIDE
 - If defined, uses  board_target_flash_chip_erase() function, instead of the default target_flash_chip_erase().
 - This function is provided in source/board/microbit/microbit_target_flash.c, and preserves two pages across erase.

 - if the BOARD_VFS_ADD_FILES macro is set, vfs_user_build_filesystem() now calls board_vfs_add_files()
 - This is defined in microbit_vfs_user.c, and registers with the VFS:
    - read_microbit_flash_bin() - read callback for binary dump of the filesystem
    - read_microbit_flash_js() - read callback for javascript array representation of the filesystem
    - read_file_microbit_html() - read callback for the html webpage
 - The location of the filesystem is not known at compile-time, and depends on how much free space is available on the NRF51822. It's location is saved in a key-value store at a known location, this is retrieved by microbit_get_flash_meta(), again in microbit_vfs_user.c

I couldn't find any docs on how to go about changing these parts of the code (?), so I've tried to keep modifications to the existing files to the minimum, and localize changes in the source/board/microbit directory. Happy to hear feedback on cleaner ways to integrate this functionality. For example, would it be better to look at the weakly defined flash_intf_target_custom instead?

### To-do
 - The HTML webpage still needs to be finalized.
 - **BUG:** currently, the functions which read the NRF flash over SWD call target_set_state(), resetting the microbit whenever it's plugged in, or a file is read. Can the NRF flash be read over SWD without putting it in reset?
 - Add a microbit-erase virtual file so that, when opened, the microbit can be erased on-demand.
 - Code review/comments
 - Testing.

Thanks,

Alex


source/daplink/interface/target_flash.c
```diff
@@ -45,7 +45,11 @@ static const flash_intf_t flash_intf = {
     target_flash_uninit,
     target_flash_program_page,
     target_flash_erase_sector,
+#ifdef TARGET_FLASH_ERASE_CHIP_OVERRIDE
+    board_target_flash_erase_chip,
```
c1728p9: I would recommend using flash_intf_target_custom (as you mentioned earlier) instead of this.  That way this code is completely contained within the microbit board folder.

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -111,6 +111,15 @@ void vfs_user_build_filesystem()
         file_size = get_file_size(read_file_need_bl_txt);
         vfs_create_file("NEED_BL TXT", read_file_need_bl_txt, 0, file_size);
     }
+    
+#ifdef BOARD_VFS_ADD_FILES
```
c1728p9: I would recommend using a weak reference for board_vfs_add_files rather than a define.

c1728p9:
Code looks good and well contained within the microbit folder.  Some minor comments on the PR.  Its cool to finally see some of these changes that I heard were coming.  As for reading flash without resetting, I had an experimental patch to do this.  It can be found here:
https://github.com/c1728p9/DAPLink/commit/77fce694645ad916e17407d2387c19fb0def1e66
I would recommend halting the microbit, reading the desired rom and then resuming.

Also, what are your thoughts on squashing these changes down to a single commit on top of master or rebasing them onto master?  If would make this change set much clearer to developers looking back on this.

c1728p9:
Hi @alexk2k12, have you had a chance to look at my feedback?

c1728p9:
@alexk2k12, @jamesadevine any updates on this?

jamesadevine:
@c1728p9 I will send you an email later on this evening!

c1728p9:
After discussion with @jamesadevine, I'm closing this change.  Feel free to re-open if the becomes useful in the future.

PR #97, opened 2016-04-30
Fix K20DX uart race condition

c1728p9:
If a uart interrupt comes in at the right time then
write_buffer.cnt_out can get incremented twice.  This causes the TX
interrupt to get stuck in a state where it is constantly transmitting
data.

This patch fixes that race condition by incrementing
write_buffer.cnt_out in just one place - the uart interrupt.

### DIFF
--- a/source/hic_hal/freescale/k20dx/uart.c
+++ b/source/hic_hal/freescale/k20dx/uart.c
@@ -23,6 +23,8 @@
 
 #include "MK20D5.h"
 #include "uart.h"
+#include "util.h"
+#include "cortex_m.h"
 
 extern uint32_t SystemCoreClock;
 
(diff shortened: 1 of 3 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
Updated this PR to fix the KL26Z which has this problem as well.  Also added a test to catch future regressions.

The bug causing the echo test does not effect the atsam3u2c and lpc11u35 due to how the TX interrupt is enabled/disabled.  The race condition causing TX count to get corrupted could not be reproduced in the atsam3u2c and lpc11u35. 

This  PR does not comprehensively fix UART race conditions, since that would require an overhaul of each uart driver.  Instead these are point fixes.

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

jupe:
this might fix some problems what we have seen..

c1728p9:
@mbed-bot: TEST

c1728p9:
This should fix #115

PR #111, opened 2016-06-15
RBL : Add software reset function for MK20DX

zoujixing:
Just for RedBearLab nRF51822 and Nano. Fixes bug "Pressing button can't reset target".

### DIFF
--- a/source/hic_hal/freescale/k20dx/gpio.c
+++ b/source/hic_hal/freescale/k20dx/gpio.c
@@ -98,7 +98,33 @@ void gpio_set_msc_led(gpio_led_state_t state)
 
 uint8_t gpio_get_sw_reset(void)
 {  
-    return (PIN_nRESET_GPIO->PDIR & PIN_nRESET) ? 1 : 0;
+    static uint8_t last_reset_forward_pressed = 0;
+    uint8_t reset_forward_pressed;
+    uint8_t reset_pressed;
+    uint32_t pcr_value;
+    // save the configuration of PIN_nRESET
+    pcr_value = PIN_nRESET_PORT->PCR[PIN_nRESET_BIT];    
+    // configure pin as GPIO input_pullup
+    PIN_nRESET_PORT->PCR[PIN_nRESET_BIT] = PORT_PCR_MUX(1) | PORT_PCR_PE_MASK | PORT_PCR_PS_MASK;
+    PIN_nRESET_GPIO->PDDR &= ~PIN_nRESET;
+    busy_wait(5);
+    reset_forward_pressed = (PIN_nRESET_GPIO->PDIR & PIN_nRESET) ? 0 : 1;
+    if(last_reset_forward_pressed != reset_forward_pressed) {
+        if(reset_forward_pressed) {
+            target_set_state(RESET_HOLD);
+        }
+        else {
+            target_set_state(RESET_RUN);
+        }
+        last_reset_forward_pressed = reset_forward_pressed;
+    }
+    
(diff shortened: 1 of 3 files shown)


source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -55,7 +55,11 @@ void gpio_init(void)
     LED_CONNECTED_GPIO->PCOR  |= 1UL << LED_CONNECTED_BIT;
     // reset button configured as gpio input
     PIN_nRESET_GPIO->PDDR &= ~PIN_nRESET;
+#if defined(RBL_nRF51)
+    PIN_nRESET_PORT->PCR[PIN_nRESET_BIT] = PORT_PCR_MUX(1) | PORT_PCR_PE_MASK | PORT_PCR_PS_MASK;
```
c1728p9: Adding a pullup to PIN_nRESET_PORT should be safe for all targets, so the "#if (RBL_nRF51)" can be removed here

c1728p9:
Please accept the contributors agreement:
https://developer.mbed.org/contributor_agreement/

c1728p9:
So software reset isn't properly supported for nRF51 targets on the k20dx?

zoujixing:
nRF51's reset pin and swclk are the same pin. RedBearLab board's button not connect to the nRF51's reset pin.

zoujixing:
I have accepted the Contributor Agreement on Mon 01 Aug 2016.

source/hic_hal/freescale/k20dx/gpio.c
c1728p9: Looks like the RBL nRF51 has a hardware pullup on the reset button, so this line isn't needed.

source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -98,7 +102,24 @@ void gpio_set_msc_led(gpio_led_state_t state)
...
+    PIN_nRESET_GPIO->PDDR &= ~PIN_nRESET;
+    busy_wait(5);
+    status = (PIN_nRESET_GPIO->PDIR & PIN_nRESET) ? 1 : 0;
+    // restore the configuration
+    PIN_nRESET_GPIO->PDDR |= 1 << PIN_nRESET_BIT;
+    PIN_nRESET_PORT->PCR[PIN_nRESET_BIT] = pcr_value;
+    
+    return status;
+#else
```
c1728p9: This code shouldn't be needed, since the RBL_nrf51 already has a pullup on the reset button according to this schematic:
http://redbearlab.com/nrf51822

Are you seeing problems without this code? If so can, you send me details on how to reproduce it?

c1728p9:
Hi @zoujixing, thanks for the PR.  The changes in gpio.c are specific to the RBL_nrf51, but could apply to any targets needing a custom reset handler. This should be made more generic so other boards and targets can take advantage of it. The lpc11u35 interface mcu already generically handles this case with some extra logic in gpio_get_sw_reset(). A similar change would be good to make here.

source/hic_hal/freescale/k20dx/gpio.c
zoujixing: In function "DAP_SETUP()", nRESET pin is configured to output, I can't get PIN_nRESET's input state when it's configured to putput.

c1728p9:
Hi @zoujixing, have you tested this patch on boards using hardware reset? Also, how critical is it that this change makes it into the upcoming release? I'll be making the release tomorrow, so this needs to be ready to go (rebased on master) and tested on at least one platform with hardware reset before the release.

0xc0170:
This needs  rebase. Sorry for taking this PR so long but some reviews were not answered.

I am closing this issue, please reopen with an update.

PR #112, opened 2016-06-18
Increase FS size to 64MB and use FAT16

c1728p9:
To allow space for larger devices increase the filesystem to 64MB.
To do this the filesystem has been switched to FAT16.


c1728p9:
@mbed-bot: TEST

Issue #94, opened 2016-04-20
If USB reset occurs in the middle of SCSI command mass storage stops working


c1728p9:
The USB stack does not propagate a reset to the endpoints.  Because of this if a USB bus reset occurs in the middle of a SCSI transaction the device will stop responding to mass storage commands.  This persists until the board is reset.

sukidog:
Curious. Under what circumstances does a USB reset occur in normal use cases? Or is this something the automated tests surfaced?

c1728p9:
This occurred after a couple of days of running automated tests every hour.  This shouldn't happen under normal usb operation.

VeliMattiLahtela:
We might have faced a similar situation during one of our automated runs:

What kind of symptoms are visible from the systems point of view?

Currently we have one device that is visible, and the serial interface can be accessed but the mount point is missing.

From dmesg we noticed this:

```
[18328.585149] VFS: busy inodes on changed media or resized disk sdc
[18328.617387] FAT-fs (sdc): unable to read boot sector to mark fs as dirty
[18328.620795] sdc: detected capacity change from 8454144 to 0
[18330.631344] sd 7:0:0:0: [sdc] 16512 512-byte logical blocks: (8.45 MB/8.06 MiB)
```

And sdc does not have a mounting point on the system.

PR #120, opened 2016-07-08
Reset MSC endpoint on USB reset to fix bug

c1728p9:
If a USB reset occurs in the middle of a SCSI command
mass storage can get in a state where it thinks the transfer
is still ongoing.  This cases new SCSI commands to get ignored
by the device.  This causes the PC to disable the MSC endpoint.

This patch fixes that problem by properly resetting the MSC endpoint
when a USB reset occurs.

### DIFF
--- a/source/usb/msc/usbd_msc.c
+++ b/source/usb/msc/usbd_msc.c
@@ -1101,6 +1101,15 @@ void USBD_MSC_BulkOut(void)
     }
 }
 
+/** \brief  Handle Reset Events
+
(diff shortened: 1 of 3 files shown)


c1728p9:
This fixes #94

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
Rebased on top of master.  There were no conflicts

PR #121, opened 2016-07-11
Added support for these boards from Embedded Artists:

embeddedartists:
LPC4088 Display Module (LPC4088DM)
  LPC4088 QuickStart Board (LPC4088QSB)


source/hic_hal/nxp/lpc11u35/gpio.c
```diff
@@ -94,11 +94,13 @@ void gpio_init(void)
     PIN_CDC_LED_IOCON = PIN_CDC_LED_IOCON_INIT;
     LPC_GPIO->SET[PIN_CDC_LED_PORT] = PIN_CDC_LED;
     LPC_GPIO->DIR[PIN_CDC_LED_PORT] |= PIN_CDC_LED;
+#ifndef GPIO_RESET_IN_NOT_SUPPORTED
```
c1728p9: This is common code so this define should be removed.  Instead, the IO_Config_Override.h mechanism should be used.

c1728p9:
Please accept the contributors agreement:
https://developer.mbed.org/contributor_agreement/

source/target/nxp/lpc4088/flash_blob.c
```diff
@@ -0,0 +1,78 @@
...
+    0x41f0e92d, 0x46054f33, 0x4614444f, 0x4a326878, 0xf0101c40, 0x60780f01, 0x5000f44f, 0x61d0bf14, 
+    0xf1b56190, 0xd3055f20, 0xf1a54622, 0xe8bd5020, 0xe6bc41f0, 0x2f00f5b5, 0x4622d305, 0x2000f5a5, 
+    0x41f0e8bd, 0xb975e6b3, 0x0100e9d4, 0xe9d44408, 0x44111202, 0x69214408, 0x69614408, 0x69a14408, 
+    0x42404408, 0x0b2861e0, 0xbf242810, 0xeb01210e, 0x213200d0, 0xf8df4e1e, 0x444e807c, 0x1000e9c6, 
+    0xf10660b0, 0x46300114, 0x697047c0, 0x2033b988, 0x0500e9c6, 0x7000f44f, 0x4002e9c6, 0x613068b8, 
+    0x0114f106, 0x47c04630, 0x28006970, 0xe8bdbf08, 0x200181f0, 0x81f0e8bd, 0x1040eb01, 0xeb01490e, 
+    0x68010080, 0x0107f021, 0x68016001, 0x60014311, 0x00004770, 0x00000004, 0x20098000, 0x000000b4, 
+    0x400fc000, 0x1fff1ff8, 0xcccccccd, 0x00000034, 0x00000014, 0x1fff1ff1, 0x4002c000, 0x00000000, 
+    0x00000001, 0x00000000, 0x00000000, 0x00000000, 
+};
```
c1728p9: Looking at the code in the FlashAlgo repo, I don't think this supports programming the external QSPI.  Is your intent to have daplink support programming more than the 512k internal rom?

source/target/nxp/lpc4088/target_lpc4088dm.c
```diff
@@ -0,0 +1,36 @@
...
+
+// The file flash_blob.c must only be included in target.c
+#include "flash_blob.c"
+
+// The LPC4088 Display Module has 512K internal flash and 16M external QSPI flash
+const target_cfg_t target_device = {
+	.sector_size    = KB(4),
+	.sector_cnt     = ((MB(16) + KB(512)) / KB(4)),
+	.flash_start    = 0,
+	.flash_end      = MB(16) + KB(512),
```
c1728p9: Does this ROM come in one block from address 0, or is that just the first 512KB?  If not would keep the 16MB out of here, as DAPLink cannot handle this currently.  Also, if you you update this for just 512KB support I would recommend combining it with target_lpc4088qsb since it will no longer be board specific and can be shared.

c1728p9: I just noticed that the flash algo has support for this:
```
    } else if (adr >= 0x80000) {
        /* This happens when a combined binary is flashed. The combined binary starts
           with 512kB that goes into the internal flash. After the 512kB comes the
           data to be written at the start of the SPIFI. SPIFI is erased on a
           need-to basis and not here. */
        return saveInSpifi(adr - 0x80000, sz, buf);
    }
```

This should work, so ignore my previous comment.

c1728p9:
@mbed-bot: TEST

PR #127, opened 2016-07-18
Relocate config admin region

flit:
Moved config admin region to below the config user region. This relieves some code size pressure on the K20DX128 bootloader, which was only a few hundred bytes from being too large to fit.


c1728p9:
Is there any other way this space can be saved?  Addresses above 0x8000 are intended to be reserved for the application and could be unintentionally erased.  Maybe some of the error messages in error.c could be switched out in bootloader mode, or maybe asserts could be updated for the bootloader to not include filename info.

flit:
Not sure… I think I understand what you mean (saving space used for strings), but will need to look at it to see how much.

Either way, we were running out of room for the K20 bootloader. Any change could have caused it to be too large. Just happened that updating the KSDK flash driver did it.

flit:
@c1728p9, Removing the filenames from asserts brings the K20 bootloader code size down to 29892 bytes (0x74c4). This leaves 1852 bytes to grow before running into the config admin region at 0x7c00.

This is what I changed the `util_assert` macro definition to:
`````{c}
#if defined(DAPLINK_IF)
// Interface assert, with the filename enabled.
#define util_assert(expression) _util_assert((expression), __FILE__, __LINE__)
#else
// Bootloader assert, filename disabled to save code size.
#define util_assert(expression) _util_assert((expression), "(file)", __LINE__)
#endif
`````

Is this ok? Shall I create a PR with this change?

c1728p9:
Yeah, I would prefer that over relocating the admin config region.

flit:
@c1728p9 Ok, will create the PR. Thanks.

c1728p9:
@flit, since this will be done in the new PR shall this be closed?

PR #114, opened 2016-06-21
Update testing to allow same fw on several boards

c1728p9:
Update the testing framework to properly handle firmware that can
change board IDs based on the board it is loaded on.  To allow this
the different configurations to be tested are hard-coded in info.py.
This allows boards, targets and firmware to be decoupled.

PR #132, opened 2016-07-29
Add target GR-PEACH

d-kato:
- Add the process for the Cortex-A
It branches in processing for Cortex-A by macro TARGET_MCU_CORTEX_A.
A "swd_host.c" for Cortex-A is created as "swd_host_ca.c".
The binary inspection was made comparison in 3 bytes of head of the data because there was no NVIC in Cortex-A.
Comparative data can be set for each board by macro BINARY_DETECTION.

- Add the partial erase processing for the chip erase is slow device.
Our Flash will take more than 20 seconds to chip erase.
Therefore, we have added the process of writing a program with partial erase.
Selective erase becomes effective by macro FLASH_PARTIAL_ERASE.

- Add target Renesas GR-PEACH
A flash of rza1h is outside the chip, and the flash and port may be different from every board.
So it was made the following path.
"/source/target/renesas/rza1h/gr-peach/"


c1728p9:
Hi @d-kato thanks for the PR. Would it be possible to move the partial erase into a second PR to make testing and merging easier?

source/daplink/drag-n-drop/flash_manager.c
```diff
@@ -95,6 +95,10 @@ error_t flash_manager_init(const flash_intf_t *flash_intf)
         return status;
     }
 
+#ifdef FLASH_PARTIAL_ERASE
+    // Processing for low-speed devices of chip erasure.
+    flash_manager_printf("    skip intf->erase_chip\r\n");
+#else
```
c1728p9: FLASH_PARTIAL_ERASE is a very useful feature. This would be nice as a run-time configurable value. Maybe a function to set the mode?
flash_manager_init(const flash_intf_t *flash_intf, flash_mode_t)
or maybe a seperate function
flash_manager_mode(CHIP_ERASE) / flash_manager_mode(SECTOR_ERASE)

What do you think @d-kato?

source/daplink/drag-n-drop/flash_manager.c
```diff
@@ -145,6 +150,18 @@ error_t flash_manager_data(uint32_t addr, const uint8_t *data, uint32_t size)
...
+            if ((current_write_block_addr % intf->erase_sector_size(current_write_block_addr)) == 0) {
+                uint32_t wk_sector = current_write_block_addr / intf->erase_sector_size(current_write_block_addr);
+                status = intf->erase_sector(wk_sector);
+                flash_manager_printf("    intf->erase_sector(sector=0x%x) ret=%i\r\n", wk_sector, status);
+                if (ERROR_SUCCESS != status) {
+                    state = STATE_ERROR;
+                    return status;
+                }
+            }
+#endif
```
c1728p9: This could cause problems if the data doesn't start on a sector boundary, which could happen during hex file programming. A safer place to put this code would be inside setup_next_sector. Same for the FLASH_PARTIAL_ERASE block below.

c1728p9: Please remove the old comment ~193 which reads "//TODO - future improvement - erase sector here" since you are implementing this :).

test/info.py
```diff
@@ -60,6 +60,7 @@
     'lpc11u35_arm_watch_efm32_if': 0x5051,
     'lpc11u35_arm_watch_nrf51_if': 0x5052,
     'lpc11u35_archlink_if': 0x9013,
+    'lpc11u35_gr_peach_if': 0x5500,
```
c1728p9: This will need to be updated since #114 was merged.

source/daplink/validation.c
```diff
@@ -23,6 +23,34 @@
...
+uint8_t validate_bin_nvic(const uint8_t *buf)
+{
+    // Very dirty hacking here for ARMv7-A (non Cortex-M) binary detection
+#ifdef BINARY_DETECTION
+    // This returns validated result when start instrunction
+    // of the buffer is BINARY_DETECTION (LDR  PC, Label)
+    // Compared with the high-order 3byte
+    if ((buf[1] == ((BINARY_DETECTION >> 8)  & 0xFF))
+     && (buf[2] == ((BINARY_DETECTION >> 16) & 0xFF))
+     && (buf[3] == ((BINARY_DETECTION >> 24) & 0xFF))) {
```
c1728p9: This should probably be overridden at a board or target level with a weakref rather than with an #ifdef.

source/daplink/drag-n-drop/flash_manager.c
d-kato: I think the "flash_manager_init()" is preferred.
In the case of "flash_manager_mode()", it might appear to be able to change at any time.

d-kato:
Hi @c1728p9 ,
Thank you very much for your comment.
Please check the change.

c1728p9:
Hi @d-kato unfortunately, I haven't had time to test this so it won't make it into 0242. I'll try to get to this by the next release. I very much want Cortex A / Peach support in DAPLink so thanks for the PR.

d-kato:
I close this PR for rebase.

PR #147, opened 2016-09-09
sam3u2c - return real free write size in uart buf

c1728p9:
Update the hal implementation of the sam3u2c to return the actual number of bytes free in the CDC buffer. Previous behavior returned that there was no space in the buffer if there were any bytes in it.


c1728p9:
@mbed-bot: TEST

sg-:
LGTM

Issue #162, opened 2016-09-27
"problem copying" on windows 8 and some Windows 10


wjzhang:
the DAPLink VFS will get the ERROR_ERROR_DURING_TRANSFER  or ERROR_TRANSFER_TIMEOUT error when "problem copying" message on Windows 8 and some Windows 10

wjzhang:
![problem copying](https://cloud.githubusercontent.com/assets/9528181/18868557/4196dac8-84db-11e6-9b07-5cfd940fc5db.png)

c1728p9:
Hi @wjzhang, thanks for the report. Can you provide more details? It useful to know the board you are using and the version of DAPLink (attach details.txt).

wjzhang:
failed information: "An error occurred during the transfer".
caused by one of our engineer build in dropbox. it add some attributes, when drag&drop in MSC, the Windows found it did not match. popup a dialog to continue. if press "yes", it should be working as CMSIS-DAP project. 

![image](https://cloud.githubusercontent.com/assets/9528181/19208887/ea4b07b6-8d32-11e6-9884-e8e4f6ec02ed.png)

[DETAILS.TXT](https://github.com/mbedmicro/DAPLink/files/517328/DETAILS.TXT)


[FAIL.TXT](https://github.com/mbedmicro/DAPLink/files/517327/FAIL.TXT)

c1728p9:
It looks like this is firmware for a board that is not yet supported on DAPLink. It has the id of an EFM32 Gecko board - '2000'. Is that correct? Also, could you share the location this code came from? Finally, does this problem only occur on your board or can you get this to occur on other boards?

wjzhang:
the test Hex file and firmware are used by our board support STM32F103 MCU. it's base on lpc11u35. compatible wtih https://developer.mbed.org/teams/mbed/wiki/SWDAP. 
this issue will happened on any boards if the drag&drop file add some special attributes when some software create it. like dropbox software. 
for create this issue, you just need create your own the drag&drop hex file in  dropbox sync folder.

c1728p9:
Hi @wjzhang, I have been able to reproduce this issue. I'll see if I can track down the cause.

c1728p9:
Hi @wjzhang, I was able to identify the issue and created https://github.com/mbedmicro/DAPLink/pull/165 to fix it. Please try this out and let me know if it fixes your problem.

wjzhang:
[...]

PR #165, opened 2016-10-09
Don't remount if file is deleted before data sent

c1728p9:
When a file with a '.bin' or '.hex' extension is found then the remount
timer starts. Abort the remount if the file is deleted before any data
is sent.

When copying files with attributes that do not exist on FAT16 windows
goes through the following steps:
1. Create an uninitialized file of the correct size
2. Delete the file created in step 1
3. Create new file with size 0
4. Prompt user:
    "Are you sure you want to copy this file without its properties"
5. If yes then send file data, if no the delete file

This causes problems because after the file in step 1 is deleted
DAPLink detects this as a transfer error and remounts. By aborting
the transfer instead of triggering an error files with incompatible
attributes can be properly copied.

This problem can be reproduced by saving a target binary to Dropbox's
sync folder and then from there copying it to DAPLink.

### DIFF
--- a/source/daplink/drag-n-drop/vfs_manager.c
+++ b/source/daplink/drag-n-drop/vfs_manager.c
@@ -140,8 +140,10 @@ static void build_filesystem(void);
 static void file_change_handler(const vfs_filename_t filename, vfs_file_change_t change, vfs_file_t file, vfs_file_t new_file_data);
 static void file_data_handler(uint32_t sector, const uint8_t *buf, uint32_t num_of_sectors);
 static bool ready_for_state_change(void);
+static void abort_remount(void);
 
 static void transfer_update_file_info(vfs_file_t file, uint32_t start_sector, uint32_t size, stream_type_t stream);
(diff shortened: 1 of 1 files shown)


c1728p9:
[daplink_log.txt](https://github.com/mbedmicro/DAPLink/files/518020/daplink_log.txt)

Log of this occurring on windows for future reference

c1728p9:
@mbed-bot: TEST

c1728p9:
This should fix #162

source/daplink/drag-n-drop/vfs_manager.c
```diff
@@ -174,6 +175,18 @@ void vfs_mngr_fs_remount(void)
     sync_unlock();
 }
 
+void vfs_mngr_fs_remount_abort(void)
```
c1728p9: This could cause problems if called externally so it should have file static scope.

c1728p9:
@mbed-bot: TEST

sg-:
Lgtm

PR #172, opened 2016-10-19
[Test] Adding delay for serial port initialization.

kgills:
I'm working on a new target and the serial port is taking some extra time to initialize. Without any delay I get the following exception when the test tries to read from the CDC_ACM class.

> serial.serialutil.SerialException: device reports readiness to read but returned no data (device disconnected or multiple access on port?

The tests executing before the serial test are resetting the device and the serial test is executing before the HIC serial port is ready.

### DIFF
--- a/test/serial_test.py
+++ b/test/serial_test.py
@@ -23,6 +23,7 @@ import functools
 import serial
 import threading
 import time
+from time import sleep
 
 ERROR_TIMEOUT_SECONDS = 10.0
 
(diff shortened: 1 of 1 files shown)


c1728p9:
What target/board is this for? Could you provide steps so I can reproduce this problem?

kgills:
I'm working on the MAX32625MBED, MAX32625HSP, and MAX32625NEXPAQ. Not sure if we've sent any to you guys yet.

kgills:
Ended up being a problem with my computer.

PR #177, opened 2016-10-31
Fix lpc11u35 lockup in USBD_WriteEP

c1728p9:
If USBD_WriteEP is called before the BUF_ACTIVE bit is cleared in the current endpoint then the code will get stuck in an infinite loop. This is because "ptr" is not volatile so the value it is pointed to is only
read once. This patch fixes that infinite loop by making the "ptr" variable volatile.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #183, opened 2016-11-10
[Flash] Make the drag-n-drop do a sector erase instead of mass.

kgills:
The drag-n-drop programming function currently does a mass erase when re-programming a target device. Having it do a sector erase will prevent bootloader/ROM code already in the flash from being erased.

### DIFF
--- a/source/daplink/drag-n-drop/flash_manager.c
+++ b/source/daplink/drag-n-drop/flash_manager.c
@@ -49,6 +49,7 @@ __attribute__((aligned(4)))
 static uint8_t buf[1024];
 static bool buf_empty;
 static bool current_sector_valid;
+static bool page_erase_enabled = false;
 static uint32_t current_write_block_addr;
 static uint32_t current_write_block_size;
 static uint32_t current_sector_addr;
@@ -95,13 +96,15 @@ error_t flash_manager_init(const flash_intf_t *flash_intf)
         return status;
     }
 
-    // Erase flash and unint if there are errors
-    status = intf->erase_chip();
-    flash_manager_printf("    intf->erase_chip ret=%i\r\n", status);
+    if (!page_erase_enabled) {
+        // Erase flash and unint if there are errors
+        status = intf->erase_chip();
+        flash_manager_printf("    intf->erase_chip ret=%i\r\n", status);
 
-    if (ERROR_SUCCESS != status) {
-        intf->uninit();
-        return status;
+        if (ERROR_SUCCESS != status) {
+            intf->uninit();
+            return status;
+        }
     }
 
     state = STATE_OPEN;
@@ -132,13 +135,13 @@ error_t flash_manager_data(uint32_t addr, const uint8_t *data, uint32_t size)
 
     // Setup the current sector if it is not setup already
     if (!current_sector_valid) {
-        current_sector_valid = true;
         status = setup_next_sector(addr);
 
         if (ERROR_SUCCESS != status) {
             state = STATE_ERROR;
             return status;
         }
+        current_sector_valid = true;
     }
 
     while (true) {
@@ -172,8 +175,6 @@ error_t flash_manager_data(uint32_t addr, const uint8_t *data, uint32_t size)
                 state = STATE_ERROR;
                 return status;
             }
-
-            //TODO - future improvement - erase sector here
         }
 
         // write buffer
@@ -237,6 +238,11 @@ error_t flash_manager_uninit(void)
     return ERROR_SUCCESS;
 }
 
+void flash_manager_set_page_erase(bool enabled)
+{
+    page_erase_enabled = enabled;
+}
+
 static bool flash_intf_valid(const flash_intf_t *flash_intf)
 {
     // Check for all requried members
@@ -275,6 +281,7 @@ static error_t setup_next_sector(uint32_t addr)
 {
     uint32_t min_prog_size;
     uint32_t sector_size;
+    error_t status;
     min_prog_size = intf->program_page_min_size(addr);
     sector_size = intf->erase_sector_size(addr);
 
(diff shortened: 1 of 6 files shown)


OSHChip-Owner:
How will this change affect environments that require a full chip erase, such as nRF51822 and nRF52832.
I think that on these chips, once the softdevice has executed, it locks that part of flash, and only a full chip erase can erase the lower end of flash where the softdevice resides.

I note that in your updated flash_manager.c at line 98, the test of status from the code you deleted (old version 98..101) is still there.

kgills:
So the nRF application locks the flash, and a mass erase can erase the protected regions, but a sector erase can not?

Removed the status code.

OSHChip-Owner:
The nRF  environment can include an optional boot loader, an optional softdevice (the Bluetooth Low Energy (BLE) software stack provided as a HEX file from Nordic) , and the application code. Usually these components are combined into one HEX file for the first time the chip is programmed. The first time the program runs on the chip, the softdevice will make the additional changes to some special registers to protect itself (from, for example the application program accidentally writing over the softdevice) by restricting the allowable operations on that part of flash. I think the restrictions are execute only, and no sector erase, only full chip erase.

If updates to the application area done, this is not a problem, and only sector erases are needed. The issue I am thinking of arises if an update to the softdevice is being attempted, or a new application that does not include a softdevice in its image, and loads to the area that the softdevice currently resides.
I think the order of these sections, starting at address 0 is vector table, bootloader, softdevice, application program.

kgills:
Ok that makes sense, but I think that relying on the mass erase is kind of a shortcoming. Seems to me that putting a chip erase, or some code to unlock those particular sections the softdevice resides in, in the target_unlock_sequence() would be more appropriate. 

```
uint8_t target_unlock_sequence(void)
{
    // if the device can secure the flash and there is a way to
    //  erase all it should be implemented here.

    if(flash_intf_target->erase_chip() != ERROR_SUCCESS) {
        return 0;
    }

    return 1;
}

```


The issue I'm trying to solve is that the target that I'm working with has something like a softdevice that we don't want to include in the application image.

OSHChip-Owner:
From a product security point of view, this one way (once you set the protection, you can't un-set it) is quite common for processors from many vendors. Mandating a full chip erase as the only way to load new firmware is part of the way you avoid various protection attack scenarios. Anyway, for Nordic, that is how it is, so maybe the solution for your issue should be in code that is specific to that processor, and not something that affects all processors that DAPLink will be used with.

kgills:
Is there anything stopping the user from sector erasing from the debugger? I understand the security concern but it seems that it would be pretty easy to work around if someone wanted to. Again I feel that if targets are relying on a mass erase to unlock the flash then that should be in the target_unlock_sequence function.

I would make a target specific change to work around this if I could, but short of buffering the data on the DAPLink micro I can't undo a mass erase.

I could use a macro to switch between mass erase or sector erase. Something to the effect of...

```
#ifndef DRAG_DROP_SECTOR_ERASE
// Do a mass erase
#else
// Do a sector erase
#endif
```

@c1728p9 What do you think?

c1728p9:
Hi @kgills, there is definitely value in adding sector erase, but this needs to be in addition to chip erase. Chip erase is faster in some cases and required for some targets, as mentioned by @OSHChip-Owner. Ideally this would be run-time configurable, even if a give board is only ever configured to use one. I would imagine this API for this being something like flash_manager_set_page_erase(bool enabled) but this still need to be determined. Also for reference, [this patch](https://github.com/mbedmicro/DAPLink/pull/163/commits/57cc40dab2585568fac1e24458afd55a22baf7f3) adds similar functionality.

kgills:
OK that works for me. What is left to be determined? I can make a PR doing that if you want.

c1728p9:
Yeah, it would be great if you could make a PR doing this.

flit:
I agree that sector erase is really useful. It should probably be the default since it won't unexpectedly erase other flash contents. If a a device requires mass erase for some reason then the default could be changed for that device. And users could always change it via an action file if they want faster updates.

Imo, security is not a serious concern here because DAPLink is meant for development use, not production or field firmware updates. You primarily just need to be able to unlock a secured device automatically or at least easily. However, it's not uncommon to have libraries such as the NXP motor control libs or Nordic softdevices in protected flash sectors, requiring you to use sector erase to not wipe them out.

kgills:
I updated the PR to give us the runtime option of enabling the sector erase.

source/daplink/drag-n-drop/iap_flash_intf.c
```diff
@@ -244,11 +244,10 @@ static error_t program_page(uint32_t addr, const uint8_t *buf, uint32_t size)
     return ERROR_SUCCESS;
 }
 
-static error_t erase_sector(uint32_t addr)
+static error_t erase_sector(uint32_t sector)
```
kgills: The definition at the top of the file uses a sector number instead of address. I ran into this issue because this erase_sector function was taking an address and the target_flash_erase_sector function was taking a sector number.

source/daplink/interface/target_flash.c
```diff
@@ -118,11 +118,16 @@ static error_t target_flash_program_page(uint32_t addr, const uint8_t *buf, uint
     return ERROR_SUCCESS;
 }
 
-static error_t target_flash_erase_sector(uint32_t sector)
+static error_t target_flash_erase_sector(uint32_t addr)
 {
     const program_target_t *const flash = target_device.flash_algo;
 
-    if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_sector, sector * target_device.sector_size, 0, 0, 0)) {
+    // Check to make sure the address is on a sector boundary
```
c1728p9: It might be good to assert here as well since this is more likely a logic error rather than a runtime error.

source/daplink/drag-n-drop/flash_manager.c
```diff
@@ -237,6 +238,11 @@ error_t flash_manager_uninit(void)
     return ERROR_SUCCESS;
 }
 
+void flash_manager_set_page_erase(bool enabled)
+{
+    page_erase_enabled = enabled;
```
c1728p9: Might want to assert that the state is STATE_CLOSED since this could have unintended side effects if changed in the middle of programming.

source/daplink/drag-n-drop/flash_manager.c
```diff
@@ -294,6 +301,17 @@ static error_t setup_next_sector(uint32_t addr)
     current_sector_size = sector_size;
     current_write_block_addr = current_sector_addr;
     current_write_block_size = MIN(sector_size, sizeof(buf));
+
+    if(page_erase_enabled) {
```
c1728p9: nit - no space between 'if' and '(page_...'

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

d-kato:
I would like to use "flash_manager_set_page_erase(true)".
Please let me know if there is a good place to call this function.

c1728p9:
I would recommend setting this in prerun_board_config

d-kato:
Hi @c1728p9, thank you. I will implement it in that way.

PR #187, opened 2016-11-14
K20DX128 HIC pin fixes

flit:
- Enable high drive strength for SWD pins on the K20DX128 HIC.
- Fix problem with nRESET being driven when SWD was disabled on the K20DX128 HIC. This resolves a long standing problem where you couldn't use DAPLink for the virtual serial port while simultaneously using another debug probe connected to the target SWD header.


c1728p9:
@mbed-bot: TEST

PR #194, opened 2016-11-29
Fix erroneous read of extra word in swd_read_block()

flit:
`swd_read_block()` was always causing the AP to issue a read of an extra word.

This is the root cause of the problem that was also addressed by #189. On Kinetis devices where the UID is read, the bug causes the word after the last UID register, an invalid register on most devices, to also be read. (#189 is still useful on its own, will continue to work on it.)

Verified that the UID is still successfully read as the same value before and after the patch.

Also replaced some literal constants in `swd_read_block()` with macros to improve readability.

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -298,32 +298,32 @@ static uint8_t swd_read_block(uint32_t address, uint8_t *data, uint32_t size)
     }
 
     // TAR write
-    req = SWD_REG_AP | SWD_REG_W | (1 << 2);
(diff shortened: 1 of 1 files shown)


c1728p9:
Thanks for the detailed investigation of this @flit.

c1728p9:
@mbed-bot: TEST

flit:
@c1728p9 Good question about Cortex-A, I'll check.

flit:
@c1728p9 `swd_host_ca.c` doesn't have a similar `swd_read_block()` function.

c1728p9:
Ah, didn't realize that.

PR #195, opened 2016-11-30
lpc11u35 - Prevent dropped USB SETUP packets

c1728p9:
When a SETUP packet comes immediately after a control transfer with a data in stage there is a chance that the OUT packet sent by the host at the end of the transfer will be dropped. This is because there is no way to tell that the SETUP and OUT are both pending so only the SETUP packet gets processed.

This patch records whether an OUT packet is expected or not on the control endpoint. If it is then it ensures that the OUT packet is processed first.

### DIFF
--- a/source/hic_hal/nxp/lpc11u35/uart.c
+++ b/source/hic_hal/nxp/lpc11u35/uart.c
@@ -40,6 +40,8 @@ static struct {
     volatile  int16_t cnt_out;
 } write_buffer, read_buffer;
 
+static int32_t reset(void);
+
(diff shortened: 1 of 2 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

sg-:
LGTM

PR #198, opened 2016-12-01
atsam3u - Prevent dropped USB SETUP packets

c1728p9:
When a SETUP packet comes immediately after a control transfer with a data in stage there is a chance that the SETUP packet will be corrupted by the OUT packet processing. This is because at the time the interrupt occurred the register UDPHS_EPTSTA contained information about the OUT packet, but in the read inside USBD_ReadEP this register contains information about the SETUP transfer that occurred. Because of this the SETUP packet is treated like an OUT packet.

This patch fixes this problem by recording the value of UDPHS_EPTSTA and using this value inside USBD_ReadEP rather than reading from the hardware register again. This ensures the value does not change the initial read and the call to USBD_ReadEP.

This problem can be reproduced by adding "os_dly_wait(1);" at the start of USBD_ReadEP before the read of UDPHS_EPTSTA. With this patch adding the delay does not effect USB stability.

### DIFF
--- a/source/hic_hal/atmel/sam3u2c/usbd_ATSAM3U2C.c
+++ b/source/hic_hal/atmel/sam3u2c/usbd_ATSAM3U2C.c
@@ -28,6 +28,7 @@
 #include "usb_config.c"
 
 #define UDPHS_EPTFIFO_BASE (0x20180000) /* (UDPHS_EPTFIFO) Base Address       */
+uint32_t eptsta_copy[USBD_EP_NUM + 1];
 
 /*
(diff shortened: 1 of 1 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

sg-:
LGTM

PR #202, opened 2016-12-12
Fix target reset problems on lpc11u35 interface

c1728p9:
On some boards reset doesn't work correctly when the debug module is turned off while the target is halted. This patch fixes that problem by always turning on debug in swd_set_target_state_*.

This problem can be reproduced by programming the target on a LPC812 Xpresso. The target will not start running when the reset button is pressed. The board needs to be disconnected and reconnected for the target to run.

This problem was introduced in the commit
c074a58052873296fa9f918341cb0e06e6ab0c03 -
Added swd_uninit_debug() function. Used new function in reset routine
for nRF5x, as well as target_flash.c uninit.

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -859,10 +859,13 @@ uint8_t swd_set_target_state_hw(TARGET_RESET_STATE state)
 {
     uint32_t val;
     swd_init();
+    swd_init_debug();
 
     switch (state) {
         case RESET_HOLD:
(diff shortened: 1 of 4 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #203, opened 2016-12-20
Only disable asserts for bootloader

c1728p9:
Asserts can be disabled to save space. This should only apply to the bootloader since the interface has enough space for asserts.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

sg-:
Do asserts need to be disabled now?

c1728p9:
Earlier they were disabled for Kinetis devices so @flit could get his patches to fit. It was only meant to turn off asserts in the bootloader, but ended up turning off asserts in the interface as well. This patch turns them back on for the interface.

sg-:
LGTM 👍

PR #205, opened 2016-12-22
Fix msd assert caused by processing after xfer

c1728p9:
Stop further file processing once a mass storage transfer is finished. This fixes an assert sometimes caused when data arrives after a transfer has been completed.


c1728p9:
@mbed-bot: TEST

PR #207, opened 2016-12-24
Improve fifo overflow

c1728p9:
Update all uart drivers to use a common queue driver for TX and RX. Update overflow behavior so once DAPLink's RX buffer has filled further characters are dropped. Also add the message "<OVFL>" to the
serial data if an overflow condition occurs.

### DIFF
--- a/source/daplink/circ_buf.c
+++ b/source/daplink/circ_buf.c
@@ -1,0 +1,137 @@
+/**
+ * @file    circular_buffer.c
+ * @brief   Implementation of a circular buffer
+ *
+ * DAPLink Interface Firmware
+ * Copyright (c) 2016-2016, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "circ_buf.h"
+
+#include "cortex_m.h"
+#include "macro.h"
+#include "util.h"
+
+void circ_buf_init(circ_buf_t *circ_buf, uint8_t *buffer, uint32_t size)
+{
+    cortex_int_state_t state;
+    state = cortex_int_get_and_disable();
+
+    circ_buf->buf = buffer;
+    circ_buf->size = size;
+    circ_buf->head = 0;
+    circ_buf->tail = 0;
+
+    cortex_int_restore(state);
+}
+
+void circ_buf_push(circ_buf_t *circ_buf, uint8_t data)
+{
+    cortex_int_state_t state;
+    state = cortex_int_get_and_disable();
+
+    circ_buf->buf[circ_buf->tail] = data;
+    circ_buf->tail += 1;
+    if (circ_buf->tail >= circ_buf->size) {
+        util_assert(circ_buf->tail == circ_buf->size);
+        circ_buf->tail = 0;
+    }
+
+    // Assert no overflow
+    util_assert(circ_buf->head != circ_buf->tail);
+
+    cortex_int_restore(state);
+}
+
+uint8_t circ_buf_pop(circ_buf_t *circ_buf)
+{
+    uint8_t data;
+    cortex_int_state_t state;
+
+    state = cortex_int_get_and_disable();
+
+    // Assert buffer isn't empty
+    util_assert(circ_buf->head != circ_buf->tail);
+
+    data = circ_buf->buf[circ_buf->head];
+    circ_buf->head += 1;
+    if (circ_buf->head >= circ_buf->size) {
+        util_assert(circ_buf->head == circ_buf->size);
+        circ_buf->head = 0;
+    }
+
+    cortex_int_restore(state);
+    
+    return data;
+}
+
+uint32_t circ_buf_count_used(circ_buf_t *circ_buf)
+{
+    uint32_t cnt;
+    cortex_int_state_t state;
+
+    state = cortex_int_get_and_disable();
+
+    if (circ_buf->tail >= circ_buf->head) {
+        cnt = circ_buf->tail - circ_buf->head;
+    } else {
+        cnt = circ_buf->tail + circ_buf->size - circ_buf->head;
+    }
+
+    cortex_int_restore(state);
+    return cnt;
+}
+    
+uint32_t circ_buf_count_free(circ_buf_t *circ_buf)
+{
+    uint32_t cnt;
+    cortex_int_state_t state;
+
+    state = cortex_int_get_and_disable();
+
+    cnt = circ_buf->size - circ_buf_count_used(circ_buf) - 1;
+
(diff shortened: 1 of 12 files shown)


source/daplink/circ_buf.c
```diff
@@ -0,0 +1,135 @@
...
+    return cnt;
+}
+
+uint32_t circ_buf_write(circ_buf_t *circ_buf, const uint8_t *data, uint32_t size)
+{
+    uint32_t cnt;
+    uint32_t i;
+
+    cnt = circ_buf_count_free(circ_buf);
+    cnt = MIN(size, cnt);
```
jupe: i think this should indicate way or another when ringbuffer is full, otherwise user doesnt even know if some data is dropped.

source/daplink/circ_buf.c
```diff
@@ -0,0 +1,135 @@
...
+
+    cortex_int_restore(state);
+}
+
+void circ_buf_push(circ_buf_t *circ_buf, uint8_t data)
+{
+    cortex_int_state_t state;
+    state = cortex_int_get_and_disable();
+
+    circ_buf->buf[circ_buf->tail] = data;
```
jupe: interesting, push to tail and read from head.. normally its otherwise

source/daplink/circ_buf.c
```diff
@@ -0,0 +1,135 @@
...
+    circ_buf->tail = circ_buf->tail + 1;
```
jupe: maybe just `circ_buf->tail += 1`, same way than line 49

source/daplink/circ_buf.c
```diff
@@ -0,0 +1,135 @@
...
+    if (circ_buf->tail >= circ_buf->size) {
+        circ_buf->tail -= circ_buf->size;
```
jupe: is there reason why this cannot be just like: `circ_buf->tail = 0`. if tail goes out of bounds (shouldn't happen ever) this doesn't go back to beginning of buffer (=0).

source/daplink/circ_buf.c
jupe: if ring is full already this just allow to push more data. i think this should drop oldest data before pushing more to ring buffer.. or return failure that buffer is full, but then reader will get old data..

source/daplink/circ_buf.c
c1728p9: The number of bytes written to the buffer is returned so from that you can determine if data was dropped.

source/daplink/circ_buf.c
c1728p9: In this implementation calling circ_buf_push when the buffer is full is considered a software error. If this happens then it will trigger an assert which will show up on DAPLink's filesystem so this will not go unnoticed.

source/daplink/circ_buf.c
c1728p9: done

source/daplink/circ_buf.c
c1728p9: Good idea. I made this change and added an assert to catch out of bounds problems.

source/hic_hal/freescale/kl26z/uart.c
```diff
@@ -283,9 +237,18 @@ void UART_RX_TX_IRQHandler(void)
...
-            read_buffer.cnt_in++;
+            uint32_t free;
+            uint8_t data;
+            
+            data = UART1->D;
+            free = circ_buf_count_free(&read_buffer);
+            if (free > RX_OVRF_MSG_SIZE) {
+                circ_buf_push(&read_buffer, data);
+            } else if (RX_OVRF_MSG_SIZE == free) {
+                circ_buf_write(&read_buffer, (uint8_t*)RX_OVRF_MSG, RX_OVRF_MSG_SIZE);
```
jupe: is this msg going to host or target? sounds little dangerous to send any data even overflow happens, e.g if there is some protocol used in serial port this might causes confusion&trouble. Instead maybe we could create fail.txt file to fs with similar message which indicate that there was overflow during data transfer? protocol anyway need somekind of validation so even some data was dropped probably it is already notified by tools/application.. but at least I propose to add line feeds to message, covers situation where protocol is simple line based. then it is handled proper way at least by our test tools..

c1728p9: This message is going directly to the host PC. I thought about having this error show up in FAIL.TXT or ASSERT.TXT, but I figured this would be too intrusive, as it would cause the drive to remount to display this. Propagating this up via serial control signals might be intrusive as well, and I'm not sure of a clear way to do this. I'll add a newline to make this easier for tools to handle. This can always be changed later if it turns out to be too intrusive.

c1728p9:
Thanks for the feedback @jupe, I updated the PR to make those changes. Let me know if you see anything else.

source/hic_hal/freescale/kl26z/uart.c
flit: Fwiw, I agree with @jupe that writing the overflow message to the host is likely to cause problems with some protocols.

c1728p9: @flit do you think this behavior is worse that just dropping the data? Either way this seems likely to cause problems with some protocols. With an error message in the serial stream at least the cause of serial data loss can be identified. I'm open to suggestions on alternate ways to fix this.

One important case this helps with is in rare cases in mbed-os testing there are dropped segments of test output that cause the testing to indicate failure. It has been hard to determine if the failure was due to DAPLink dropping data due to an overflow or a real failure.

jupe: could we use proper flow controls to avoid this in first place? xon/xoff or rts/cts...

c1728p9: There are a few targets which do support and use hardware flow control, such as the nrf51-dk. Most, like the K64F don't though.

DAPLink does throttle USB CDC data coming from the host PC, so DAPLink's buffers in this direction will never overflow. The problem occurs when data is going from DAPLink to the PC. If the PC stops reading data from DAPLink then DAPLink's buffers will fill. If DAPLink can't stop the target from sending more data (as is the case in targets without flow control) then eventually DAPLink's buffers will overflow.

c1728p9: @flit, @jupe, another possibility (if no alternate mechanism can be found and the error message is too intrusive) is to only enable the overflow message with a config which is turned off by default.

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
Changed the overflow text from "<OVFL>" to "<DAPLink:Overflow>" and added a configuration option to turn on/off overflow reporting. By default overflow reporting is turned off.

c1728p9:
@mbed-bot: TEST

flit:
Having the overflow reporting as a config option is perfect.

And to answer your question, yes, I do think that if you're running some protocol then it's probably better to just drop data. Although, a well-designed protocol should be able to handle both dropped data and garbage data inserted midstream by detecting the error and retransmitting.

sg-:
LGTM

source/daplink/circ_buf.c
jupe: `When buffer overflow is happening, oldest data must be overwritten with new one. `

This doesn't fulfill original issue yet, or is it ?

c1728p9: @jupe, the original issues was that DAPLink kept a count of bytes sent by the target and tried to send that number of bytes back to the host, even if the buffer already overflowed. This gave the appearance of DAPLink getting stuck sending data to the host.

This PR addresses that, along with unifying the queuing mechanism across the targets.

jupe: I think another issue was that when buffer overflow happens daplink will send old data instead latest one when host start reading again... @terhei did I remember correctly ?

PR #212, opened 2017-01-01
Move serial processing to main thread

c1728p9:
Move processing of the CDC data to the main thread to prevent concurrent access to the CDC USB driver's data structures. This also frees up memory since the CDC mailbox and thread are no longer used.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #213, opened 2017-01-02
k20dx - Remove superficial read modify write

c1728p9:
Do not read the GPIO PCOR register before writing to it. This register always has the value zero. This makes it clear that this operation will not have side effects aside from clearing the desired bit.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #215, opened 2017-01-03
Add short startup delay to gpio_init like in k20dx port.

mazgch:
Add short delay to allow for delays between split power sources for target and debugger.

Otherwise if target power supply is late then it keeps entering the maintenance mode if NRESET_FB_ISP is not detected high.

### DIFF
--- a/source/hic_hal/freescale/kl26z/gpio.c
+++ b/source/hic_hal/freescale/kl26z/gpio.c
@@ -21,6 +21,16 @@
 
 #include "gpio.h"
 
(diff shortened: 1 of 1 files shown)


c1728p9:
Hi @mazgch which board are you seeing this problem on?

mazgch:
It's a new board which has an additional external LDO for the target and does not use the same LDO as the KL26Z interface chip. KL26Z is powered by its own internal LDO.

c1728p9:
@mbed-bot: TEST

PR #216, opened 2017-01-06
Revert "Added swd_uninit_debug() function. Used new function in reset…"

c1728p9:
This reverts commit c074a58052873296fa9f918341cb0e06e6ab0c03. This is reverted because it causes some lpc targets such as the lpc812 to enter a bad state after programming, since the DP is turned off while the debug logic is still enabled.

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -831,25 +831,6 @@ uint8_t swd_init_debug(void)
     return 1;
 }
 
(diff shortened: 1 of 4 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

sg-:
LGTM 👍 

We can re-apply once the sequence works for all targets or is implemented as device specific.

PR #218, opened 2017-01-08
Fix release package script to handle underscores

c1728p9:
The package release script incorrectly handles underscores in the base name of firmware. Instead of getting the whole name it only grabs up to the first underscore. For example for the base name of 'sam3u2c_mkit_dk_dongle_nrf5x_if' is determined to be 'mkit' rather than 'mkit_dk_dongle_nrf5x'. This patch fixes that problem by joining the middle elements of the name to create the correct base name.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #221, opened 2017-01-10
Verify data flashed when automation is enabled

c1728p9:
If automation is enabled then read back every page written to verify the contents are correct. This allows devices with worn out flash to be detected.


c1728p9:
@mbed-bot: TEST

PR #227, opened 2017-02-01
Drop oldest character on UART overflow

c1728p9:
When overflow detection is not turned on and and overflow occurs drop the oldest rather than the newest data. Newest data is still dropped if overflow detection is turned on.


c1728p9:
@mbed-bot: TEST

c1728p9:
@jupe, @terhei

jupe:
agree with tero. This logic should be in circular side.

terhei:
Ok, all logic is currently in IRQ handler, and circular buffer is only having buffer functionality. In that case, PR is OK for me.

PR #228, opened 2017-02-07
Fix FRDM-KW24D target

flit:
- Renamed FRDM-KW24F to FRDM-KW24D.
- Added missing KW24 target, there was only the totally different K24F.
- Fixed flash algo.


c1728p9:
Thanks for the update

c1728p9:
@mbed-bot: TEST

flit:
@c1728p9 Fyi, I haven't been able to test since I don't have a FRDM-KW24D board…

c1728p9:
@flit testing failed since SUPPORTED_CONFIGURATIONS wasn't updated with the new name

c1728p9:
I tried this on a board locally and found that ram_end needed to be set to 0x20008000 rather than 0x20007fff to get programming to work. I also updated info.py. I pushed these change here:
https://github.com/c1728p9/DAPLink/tree/dev_fix_kw24

flit:
So is the ram_end actuall ram end + 1? How did that cause it to fail? (How is ram_end used?)
Sorry I forgot about info.py—I always seem to forget this file…
Will update the PR.

c1728p9:
The ram size is used to validate the reset handler entry of the vector table. In the image I was testing with the reset handler was set to 0x20008000 so DAPLink didn't recognize it as a valid binary.

flit:
You mean the stack pointer I take it? I see. Should probably fix it at some point so the check handles the SP being set to the word after the last word of RAM, rather than requiring `ram_end` to be + 1. Not a big deal though.

Anyway, fixes pushed. Thanks!

c1728p9:
Thanks for the update!

c1728p9:
@mbed-bot: TEST

c1728p9:
Also, ram_end is confusing and has been the source of several errors. Maybe it would be best to have a ram_start and ram_size so any ambiguity is removed.

flit:
That sounds good.

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

PR #230, opened 2017-02-15
Prevent lockup when target DP won't powerup

c1728p9:
If for some reason a targets DP won't powerup, DAPLink will get stuck in a loop waiting for this to occur. Add a timeout to prevent this timeout and to allow DAPLink to boot normally.

This problem can be reproduced on the FRDM-KW24D if target firmware sets pin PTB19 low immediately after boot.

CC: @flit

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -782,6 +782,8 @@ static uint8_t JTAG2SWD()
 uint8_t swd_init_debug(void)
 {
     uint32_t tmp = 0;
+    int i = 0;
+    int timeout = 100;
     // init dap state with fake values
     dap_state.select = 0xffffffff;
     dap_state.csw = 0xffffffff;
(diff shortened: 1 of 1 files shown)


c1728p9:
@mbed-bot: TEST

flit:
Strange, I don't see a PTB19 pin on the KW24D pinout. Do you mean PTA19 (XTAL0)?

c1728p9:
You'll need to look at the reference manual to find it. It is an internal only pin.

flit:
Thanks, I see it now. Any idea why putting the modem into reset causes the DP to not power up?

c1728p9:
I have no idea how a processor could turn off and keep off its own DP.  This problem is easy to reproduce if you or someone else wanted to investigate it further. Oddly enough, jtag seems to work normally.
CC: @maclobdell

c1728p9:
@mbed-bot: TEST

flit:
Agreed, I don't see how that's possible. I'll ask around.

What kind of timing is required for pulling PTB19 low? Is there an amount of time after which it can be pulled low without causing this problem?

c1728p9:
Pull it low at any time and it causes problems. It needs to be pulled low on boot to cause daplink to get stuck (which this PR fixes)

flit:
I see. Does the DP return to normal if you set PTB19 high?

I asked the connectivity software team if they have ever seen debug issues caused by this pin (they use J-Links).

c1728p9:
I'm not sure if setting PTB19 high fixes this, as the micro doesn't seem to work properly once it is low. I tried programming with an external j-link and it seemed to have this same problem with swd, but not with jtag. You might ask the connectivity specifically if they have seen this with swd.

flit:
Ok, thanks.

flit:
@c1728p9 Feedback from the connectivity software team:
- If the Radio_RST is asserted after the MCU initializes the clock reference to use the CLK from the radio, this would leave the MCU with no clock.
- We noticed that the internal clock from the KW2xD is not calibrated they could try using FEI but first calibrate the internal clock using radio's clock as external reference. https://community.nxp.com/docs/DOC-333573 
- RST_B is routed to PTB19 and the SWD pins are in PTAx so the RST asserting itself should not affect the SWD/JTAG interface.

PR #231, opened 2017-02-16
Add retry logic to mass storage test

c1728p9:
Under certain conditions a PC running DAPLink tests can clear the disk cache DAPLink is mounted on. When this occurs the host OS will re-read disk contents. Since DAPLink's disk contents never change the
drive will look like it did when it first mounted. None of the new files that were created will exist. If this cache clear occurs in the middle of a write operation in the mass storage test an IOError will be raised, causing testing to terminate abruptly.

To prevent testing from spuriously failing this patch adds retry logic to the mass storage test. This prevents these spurious failures.

Note - a disk cache clear can be triggered by allocating most of the systems memory in a program. For example on a windows 10 PC with 16GB of RAM running the following code (which allocates 15GB) in a python terminal will typically trigger a cache clear of all disks:
"a = bytearray(1024 * 1024 * 1024 * 15)"

### DIFF
--- a/test/msd_test.py
+++ b/test/msd_test.py
@@ -24,6 +24,7 @@ import shutil
 import six
 import info
 import intelhex
+from test_info import TestInfo
 
 
 def _same(d1, d2):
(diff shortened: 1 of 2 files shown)


c1728p9:
@mbed-bot: TEST

PR #237, opened 2017-02-27
More usb fixes

c1728p9:
Fix a problem with the atsam3u and a generic problem causing CBW packets to get stalled if they occur at the right time.

PR #238, opened 2017-02-27
Pyusb testing

c1728p9:
Add tests at the USB level to reproduce and test for problems that can't be reliably reproduced when using the host msd,  cdc and hid drivers. Also, add a DAPLink USB test command to simulate worst case device USB driver latency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ requests
 intelhex
 six
 enum34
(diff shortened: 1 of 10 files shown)


c1728p9:
Depends on #237

c1728p9:
@mbed-bot: TEST

PR #240, opened 2017-03-06
Configurable IO on atsam3u

c1728p9:
Make the IO of the atsam3u compile time configurable to support boards with slight pinout variations.


c1728p9:
@mbed-bot: TEST

PR #242, opened 2017-03-07
Don't assert on unused targets when testing

c1728p9:
When loading a bundle of targets, don't assert if the target isn't used. This allows new targets to be added in advance before DAPLink work starts.


c1728p9:
@mbed-bot: TEST

PR #253, opened 2017-03-24
Fix serial problems

c1728p9:
Fix problems that occur when serial data is sent by the target while DAPLink is booting.


c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

jupe:
this seems to related to all targets, is it?

c1728p9:
Hi @jupe, yep, this problem is common to all interfaces. It was introduced in the 0243 release.

c1728p9:
@mbed-bot: TEST

PR #266, opened 2017-05-08
Added RO359B, board id 1022

ytsuboi:
CC @toyowata

### DIFF
--- a/test/info.py
+++ b/test/info.py
@@ -53,6 +53,7 @@ PROJECT_RELEASE_INFO = {
     ("lpc11u35_hrm1017_if",                         False,      0x0000,     "bin"       ),
     ("lpc11u35_sscity_if",                          False,      0x0000,     "bin"       ),
     ("lpc11u35_ssci824_if",                         False,      0x0000,     "bin"       ),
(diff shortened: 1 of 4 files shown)


test/info.py
```diff
@@ -101,6 +102,7 @@
     (   0x1017,     'lpc11u35_hrm1017_if',                      None,               'mbed-HRM1017'                          ),
     (   0x1018,     'lpc11u35_ssci824_if',                      None,               'Switch-Science-mbed-LPC824'            ),
     (   0x1019,     'lpc11u35_sscity_if',                       None,               'Switch-Science-mbed-TY51822r3'         ),
+    (   0x1022,     'lpc11u35_ro359b_if',                       None,               'ro359b'                                ),
```
c1728p9: If the ro359b target can't be built using the online compiler you'll need to set it to None for now. If you do this just leave a TODO to the right with it.

ytsuboi: Hi @c1728p9 , is [this](https://github.com/mbedmicro/DAPLink/pull/266/commits/040e5a2b1141934a630b612ce325f37c8298a7be) right fix?

PR #272, opened 2017-05-25
Add framework for bootloader updates

c1728p9:
Allow bootloader builds to be created by defining DAPLINK_BOOTLOADER_UPDATE. When this is defined the bootloader image will be built into the interface. When the interface runs for the first time it will update the bootloader.

### DIFF
--- a/docs/BOOTLOADER_UPDATES.md
+++ b/docs/BOOTLOADER_UPDATES.md
@@ -1,24 +1,24 @@
 # Bootloader updates
 
-DAPLink has the ability to bundle the bootloader firmware with the interface firmware and apply a bootloader update on the first boot of the interface. This allows bootloader updates to be applied as the same time as interface updates and without any special instructions for the user.
+DAPLink has the ability to bundle the bootloader firmware with the interface firmware and apply a bootloader update on the first boot of the interface. This allows bootloader updates to occur at the same time as interface updates and without any special instructions for the user.
 
 ### Enabling bootloader updates
 
-To enable bootloader updates for a board or interface circuit the value ```DAPLINK_BOOTLOADER_UPDATE``` must be defined. Once defined, the normal release process which documented on the [developers guide](DEVELOPERS-GUIDE.md) can be used to create firmware for all targets.
+To enable bootloader updates for a board or interface circuit, you must define the value ```DAPLINK_BOOTLOADER_UPDATE```. Once you define it, you can use the normal release process, which [developers guide](DEVELOPERS-GUIDE.md) documents, to create firmware for all targets.
 
 ### Safety when updating the bootloader
 
-The vector table of the bootloader is erased and replaced during the update. Because of this, there is a small window of time where if power is lost the device will be in a non-bootable state since it does not have a valid vector table. To minimize the time without a valid vector table and the reduce risk the bootloader vector table is replaced with the interface's vector table before the update is performed. This way while the bulk of the bootloader is being loaded there is a valid vector table and a loss of power event does not prevent the device from booting. The update process is shown below with the critical sections where power cannot be lost showin in bold:
-1. **Erase the boot vector table and program the intenterface's vector table to this location**
-1. Erase and program each sector of the bootloader with the new firmware
-1. **Erase the boot vector table and program the new bootloader's vector table to this location**
+The vector table of the bootloader is erased and replaced during the update. Because of this, there is a small window of time during which the device will be in a nonbootable state if the device loses power. The cause of this is the lack of a valid vector table. To minimize the time without a valid vector table and to reduce risk, the bootloader vector table is replaced with the interface's vector table before the update is performed. This way, while the bulk of the bootloader is being loaded, there is a valid vector table, and a loss of power event does not prevent the device from booting. The update process is below. The critical sections during which the device cannot lose power are in bold:
+1. **Erase the boot vector table, and program the intenterface's vector table to this location**.
+1. Erase and program each sector of the bootloader with the new firmware.
+1. **Erase the boot vector table, and program the new bootloader's vector table to this location**.
 
 Other checks
 * If the current bootloader has a valid signature and a version greater than the bootloader stored in interfacer has, the update will not be applied.
-* The interface does a checksum over itself to check for any corruption. If the interface is found to be corrupt the bootloader update will not be applied and an assert will appear on the device.
-* If the current bootloader is identical to the bootloader stored in the interface the update will not be applied.
+* The interface does a checksum over itself to check for any corruption. If it finds the interface corrupt, the bootloader update will not be applied, and an assert will appear on the device.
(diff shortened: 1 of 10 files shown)


tools/post_compute_crc.py
```diff
@@ -109,6 +112,23 @@ def main():
     with open(output_file_txt, 'wb') as file_handle:
         file_handle.write("0x%08x\r\n" % crc32)
 
+    # Write out data as a C array
+    data = new_hex_file.tobinarray(start=start, size=size)
+    data = list(bytearray(data))
+    output_data = bytearray('static const unsigned int image_start = 0x%08x;\n'
+                            'static const unsigned int image_size = 0x%08x;\n'
+                            'static const char image_data[0x%08x] = {\n    ' %
```
sg-: Are there any alignment considerations to be accounted for?

c1728p9: The image data doesn't have any alignment requirements, since the flash manager code copies potentially unaligned data to an aligned buffer in RAM before flashing.

c1728p9:
Moved ```bootloader_check_and_update``` to after ```config_init``` in the startup sequence so asserts that occur during the update are handled correctly.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+1. Erase and program each sector of the bootloader with the new firmware
+1. **Erase the boot vector table and program the new bootloader's vector table to this location**
+
+Other checks
+* If the current bootloader has a valid signature and a version greater than the bootloader stored in interfacer has, the update will not be applied.
+* The interface does a checksum over itself to check for any corruption. If the interface is found to be corrupt the bootloader update will not be applied and an assert will appear on the device.
+* If the current bootloader is identical to the bootloader stored in the interface the update will not be applied.
+
+### Dangers
+* The DAPLink bootloader has a fixed location for where the interface is loaded (typically an offset of 0x8000 from the start of rom). If the bootloader you updated from had a different offset then the interface firmware you used with that bootloader will no longer work. You must use interface firmware built for the new offset.
```
AnotherButler: Please separate the second sentence into two sentences for clarity.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+1. **Erase the boot vector table and program the intenterface's vector table to this location**
+1. Erase and program each sector of the bootloader with the new firmware
+1. **Erase the boot vector table and program the new bootloader's vector table to this location**
+
+Other checks
+* If the current bootloader has a valid signature and a version greater than the bootloader stored in interfacer has, the update will not be applied.
+* The interface does a checksum over itself to check for any corruption. If the interface is found to be corrupt the bootloader update will not be applied and an assert will appear on the device.
+* If the current bootloader is identical to the bootloader stored in the interface the update will not be applied.
```
AnotherButler: Please change "the update will not be applied" to active voice to show who or what is doing the applying.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+1. **Erase the boot vector table and program the intenterface's vector table to this location**
+1. Erase and program each sector of the bootloader with the new firmware
+1. **Erase the boot vector table and program the new bootloader's vector table to this location**
+
+Other checks
+* If the current bootloader has a valid signature and a version greater than the bootloader stored in interfacer has, the update will not be applied.
+* The interface does a checksum over itself to check for any corruption. If the interface is found to be corrupt the bootloader update will not be applied and an assert will appear on the device.
```
AnotherButler: Please change "the bootloader update will not be applied" to active voice to show who or what is doing the applying.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Safety when updating the bootloader
+
+The vector table of the bootloader is erased and replaced during the update. Because of this, there is a small window of time where if power is lost the device will be in a non-bootable state since it does not have a valid vector table. To minimize the time without a valid vector table and the reduce risk the bootloader vector table is replaced with the interface's vector table before the update is performed. This way while the bulk of the bootloader is being loaded there is a valid vector table and a loss of power event does not prevent the device from booting. The update process is shown below with the critical sections where power cannot be lost showin in bold:
+1. **Erase the boot vector table and program the intenterface's vector table to this location**
+1. Erase and program each sector of the bootloader with the new firmware
+1. **Erase the boot vector table and program the new bootloader's vector table to this location**
+
+Other checks
+* If the current bootloader has a valid signature and a version greater than the bootloader stored in interfacer has, the update will not be applied.
```
AnotherButler: - Query: Is the "and" in the first sentence supposed to be "than", or am I misreading this?
- Please change "the update will not be applied" to active voice" to show who or what is doing the applying.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Enabling bootloader updates
+
+To enable bootloader updates for a board or interface circuit the value ```DAPLINK_BOOTLOADER_UPDATE``` must be defined. Once defined, the normal release process which documented on the [developers guide](DEVELOPERS-GUIDE.md) can be used to create firmware for all targets.
+
+### Safety when updating the bootloader
+
+The vector table of the bootloader is erased and replaced during the update. Because of this, there is a small window of time where if power is lost the device will be in a non-bootable state since it does not have a valid vector table. To minimize the time without a valid vector table and the reduce risk the bootloader vector table is replaced with the interface's vector table before the update is performed. This way while the bulk of the bootloader is being loaded there is a valid vector table and a loss of power event does not prevent the device from booting. The update process is shown below with the critical sections where power cannot be lost showin in bold:
```
AnotherButler: Please change the end of the third sentence to active voice to show who or what is replacing the vector table and performing the update.

docs/BOOTLOADER_UPDATES.md
```diff
@@ -0,0 +1,24 @@
...
+1. **Erase the boot vector table and program the intenterface's vector table to this location**
+1. Erase and program each sector of the bootloader with the new firmware
+1. **Erase the boot vector table and program the new bootloader's vector table to this location**
```
AnotherButler: Query: Are 1 and 3 supposed to be the same?

c1728p9: Nope, in 1 the interface's vector table is programmed, while in 3 the bootloader's vector table is programmed.

docs/BOOTLOADER_UPDATES.md
c1728p9: Its supposed to be "and". If both condition 1 and condition 2 are met then the update will not be applied.

PR #273, opened 2017-05-26
Fix lockup on lpc11u35 when flow control is on

c1728p9:
If flow control is on on the lpc11u35 then the code can get stuck indefinitely in uart reset waiting for all the bytes to get sent. To prevent this, temporarily switch the UART to loopback mode so it can be drained even if flow control is on.


iotvietmember:
@c1728p9 I think that this commit should be merged soon.

PR #274, opened 2017-05-30
[NCS36510] Update to support dual flash banks (640K)

maclobdell:
Update flash algo and target settings to support flashing of the full 640K flash.


maclobdell:
This is now based on updated FlashAlgo from https://github.com/mbedmicro/FlashAlgo/pull/32.  It has been tested by loading a few different application images (some <320K and some >320K) and confirming they function correctly.

PR #283, opened 2017-07-12
Only copy flash algo to ram once

c1728p9:
On each flash algorithm call the flash algorithm blob is copied to ram. This is unnecessary and slows down flash programming. This patch removes the code in swd_write_debug_state responsible for this copy on every flash algo call.

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -585,12 +585,6 @@ static uint8_t swd_write_debug_state(DEBUG_STATE *state)
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


c1728p9:
NCS36510 fails with this PR due to the bug fixed here:
https://github.com/mbedmicro/FlashAlgo/pull/35

c1728p9:
Depends on this #319 to fix the ncs36510's flash algo

maclobdell:
cc @ARMmbed/onsemi-mbed-team

PR #286, opened 2017-07-19
Update readme with known windows flashing issue

deepikabhavnani:
Drag and drop fails on Windows, for the first time after updating daplink firmware, and as a result spurious failures in CI were observed.
Root cause: Windows tries to create "System Volume Information" folder in between transferring the program data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,6 @@ Look for an interesting feature or defect [under issues](https://github.com/mbed
 Information for setting up a development environment, running the tests or creating a release build [can be found in the developers guide.](docs/DEVELOPERS-GUIDE.md)
 
 ## Known Issues
-Drag-n-drop program or application flashing fails for the first time on Windows machine with failure message "File sent out of order by PC. Target might not be programmed correctly". Windows tries to create "System Volume Information" folder while sending program data, which results in out of order error. Solution to this issue is:
-* Disable creation of "System Volume Information" folder on [USB drives](http://www.thewindowsclub.com/prevent-system-volume-information-folder-usb).
+Drag-n-drop program or application flashing fails for the first time after firmware update on Windows machine with failure message "File sent out of order by PC. Target might not be programmed correctly". Windows tries to create "System Volume Information" folder while sending program data, which results in out of order error. Solution to this issue is:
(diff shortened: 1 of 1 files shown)


yoonghm:
Do you want to try this?
 [http://www.thewindowsclub.com/prevent-system-volume-information-folder-usb](http://www.thewindowsclub.com/prevent-system-volume-information-folder-usb)

deepikabhavnani:
I have tried that solution and added link to that in readme update as well

c1728p9:
Just changing this setting didn't stop system volume information from getting created for me. I had to stop the service "Storage Service" in Services.msc.

CC @brandonboesch

README.md
```diff
@@ -23,3 +23,8 @@ Look for an interesting feature or defect [under issues](https://github.com/mbed
 
 ## Develop
 Information for setting up a development environment, running the tests or creating a release build [can be found in the developers guide.](docs/DEVELOPERS-GUIDE.md)
+
+## Known Issues
+Drag-n-drop program or application flashing fails for the first time on Windows machine with failure message "File sent out of order by PC. Target might not be programmed correctly". Windows tries to create "System Volume Information" folder while sending program data, which results in out of order error. Solution to this issue is:
```
c1728p9: Does this reliably happen the first time for you? For me, it seems to happen randomly. If you plug in your board and wait does it make this less likely to happen? Also, is this only for binary files, or does this happen to hex files as well?

README.md
```diff
@@ -23,3 +23,8 @@ Look for an interesting feature or defect [under issues](https://github.com/mbed
...
+* Disable creation of "System Volume Information" folder on [USB drives](http://www.thewindowsclub.com/prevent-system-volume-information-folder-usb).
```
c1728p9: This didn't work on Windows 10 when @brandonboesch and I tried it. We had to disable storage service.

README.md
deepikabhavnani: It happens 80% of time, after flashing new Daplink code, else it is random. I have not seen this with hex files.

README.md
deepikabhavnani: Can update the doc to do both.

c1728p9: When System Volume Information is turned off, does this issue stop occurring for you?

README.md
c1728p9: This should happen just as often with hex files if the file is sent out of order. @deepikabhavnani do you still have the logs of this occurring? I would like to investigate this further. It may be something that can be improved in the DAPLink firmware, rather than requiring a user to change settings on their machine.

PR #294, opened 2017-07-26
Fix USB uframe rate on atsam3u

c1728p9:
When in high speed mode, the atsam3u hal drops start of frame SOF interrupts. This causes a MICRO_SOF rate 1/8 slower than it should be since SOF and MICRO_SOF never occur at the same time. This patch fixes that problem by processing SOF interrupts in addition to MICRO_SOF interrupts.

Note - When in USB test mode during the USB tests, the timing occasionally caused the atsam3u to only receive SOF interrupts and no MICRO_SOF interrupts for more than 10s at a time. When in this state the CDC port would stop sending UART data to the host PC, since CDC IN transfers are started on the start of frame handling which was only triggered by a MICRO_SOF interrupt. This caused an intermittent serial read timeout in the when running the all endpoint test. With the addition of processing SOF interrupts, both types of frame interrupts are handled, so frame handling can't be starved by too many of one type.


c1728p9:
This fixes more intermittent CI failures

PR #298, opened 2017-08-14
Turn on bootloader updates

c1728p9:
Turn on bootloader updates for supported devices.


c1728p9:
Closing since this was never intended for merging.

PR #299, opened 2017-08-15
Fix ".DS_Store" problem on macOS

cederom:
Verify if project path is a directory.
This fixes ".DS_Store" problem on macOS.

Signed-off-by: CeDeROM Tomasz CEDRO <tomek@cedro.info>

### DIFF
--- a/test/daplink_firmware.py
+++ b/test/daplink_firmware.py
@@ -96,9 +96,10 @@ class ProjectFirmwareBundle(firmware.FirmwareBundle):
         firmware_list = []
         for name in project_dir_list:
             build_dir = tool_dir + os.sep + name + os.sep + 'build'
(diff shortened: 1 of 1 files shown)


cederom:
This seems to fix https://github.com/mbedmicro/DAPLink/issues/301 :-)

cederom:
i have also added simple shell script to loop testing on a local machine.. do we also want it here? :-)

c1728p9:
Probably best as its own PR. I wouldn't worry about moving it now

cederom:
would like to put  more patches here.. do we prefer bundle or separate PR per issue? :-)

c1728p9:
Any changes that can stand alone should go on a separate PR.

PR #302, opened 2017-08-19
Update Python version requirement

dmsg:
mbed-ls is required by DAPLink, but the install scripts of mbed-ls doesn't support Python3 now.


cederom:
Hello @dmsg, this is correct, Python 3 is not supported at this point.. we are also working on a new more detailed documentation, it should show up soon! Thank you for your feedback! :-)

PR #304, opened 2017-08-20
Test: IF<->BL switch + remount [work in progress]

cederom:
Test framework updates related with IF/BL mode switch and re-mount mechanism.
Work In Progress / Do Not Pull Yet..


cederom:
this is supposed to fix https://github.com/mbedmicro/DAPLink/issues/300

PR #319, opened 2017-09-22
Update ncs36510 flash algo to fix erase

c1728p9:
Rebuild the ncs36510 flash algo with a fix to properly wait for mass erase to complete. This allows the flash algo to work correctly when flash init is not called before each flash operation.


c1728p9:
CC: @maclobdell

maclobdell:
cc @danclement, @jacobjohnson-ON

c1728p9:
This is built from https://github.com/mbedmicro/FlashAlgo/pull/35

PR #327, opened 2017-10-15
Disable UART HWFC

iotvietmember:
For mbed Online Compiler because the VBLUNO51 use nrf51822 target

Signed-off-by: iotvietmember <robotden@gmail.com>


c1728p9:
Hi @iotvietmember was there a problem with flow control, or is this being turned off for another reason?

iotvietmember:
Hi @c1728p9, the hardware flow control is good. Because the VBLUNO51 board must use `nrf51822` target on mbed Online Compiler that do not support hwfc feature, so the VBLUNO51 board will disable hwfc feature.

PR #329, opened 2017-10-17
K20 HIC waits until USB is configured to enable target power

flit:
Previously the K20DX128 HIC would immediately enable the target power switch in `gpio_init()`. This can cause problems with certain computers on boards (i.e., upcoming MIMX6RT1050-EVK) that have high inrush currents. This change causes the interface to wait until USB is configured to turn on the power switch.

- Added `gpio_handle_usb_connected()` to gpio.h. This is is stubbed out for all HICs other than K20.
- Interface main function calls `gpio_handle_usb_connected()` when USB is configured, and when it is about to be disconnected. The bootloader is not modified.
- The K20 HIC implements this function to enable the target power switch only after high power has been negotiated with the USB host. It also turns off the power switch if USB is being disconnected.

### DIFF
--- a/source/daplink/interface/main.c
+++ b/source/daplink/interface/main.c
@@ -57,13 +57,17 @@
 #define FLAGS_MAIN_CDC_EVENT    (1 << 11)
 // Used by msd when flashing a new binary
 #define FLAGS_LED_BLINK_30MS    (1 << 6)
+
 // Timing constants (in 90mS ticks)
-// USB busy time
+// USB busy time (~3 sec)
 #define USB_BUSY_TIME           (33)
-// Delay before a USB device connect may occur
+// Delay before a USB device connect may occur (~1 sec)
 #define USB_CONNECT_DELAY       (11)
+// Timeout for USB being configured (~2 sec)
+#define USB_CONFIGURE_TIMEOUT   (22)
 // Delay before target may be taken out of reset or reprogrammed after startup
 #define STARTUP_DELAY           (1)
+
 // Decrement to zero
 #define DECZERO(x)              (x ? --x : 0)
 
@@ -209,6 +213,7 @@ __task void main_task(void)
     gpio_led_state_t msc_led_value = GPIO_LED_OFF;
     // USB
     uint32_t usb_state_count = USB_BUSY_TIME;
+    uint32_t usb_no_config_count = USB_CONFIGURE_TIMEOUT;
     // thread running after usb connected started
     uint8_t thread_started = 0;
     // button state
@@ -270,6 +275,8 @@ __task void main_task(void)
         if (flags & FLAGS_MAIN_POWERDOWN) {
             // Disable debug
             target_set_state(NO_DEBUG);
+            // Disable board power before USB is disconnected.
+            gpio_set_board_power(false);
             // Disconnect USB
             usbd_connect(0);
             // Turn off LED
@@ -302,15 +309,18 @@ __task void main_task(void)
             switch (usb_state) {
                 case USB_DISCONNECTING:
                     usb_state = USB_DISCONNECTED;
+                    // Disable board power before USB is disconnected.
+                    gpio_set_board_power(false);
                     usbd_connect(0);
                     break;
 
                 case USB_CONNECTING:
-
                     // Wait before connecting
                     if (DECZERO(usb_state_count) == 0) {
                         usbd_connect(1);
                         usb_state = USB_CHECK_CONNECTED;
+                        // Reset connect timeout
+                        usb_no_config_count = USB_CONFIGURE_TIMEOUT;
                     }
 
                     break;
@@ -322,7 +332,16 @@ __task void main_task(void)
                             thread_started = 1;
                         }
 
+                        // Let the HIC enable power to the target now that high power has been negotiated.
+                        gpio_set_board_power(true);
+
                         usb_state = USB_CONNECTED;
+                    }
+                    else if (DECZERO(usb_no_config_count) == 0) {
+                        // USB configuration timed out, which most likely indicates that the HIC is
+                        // powered by a USB wall wart or similar power source. Go ahead and enable
+                        // board power.
+                        gpio_set_board_power(true);
                     }
 
                     break;
--- a/source/hic_hal/freescale/k20dx/gpio.c
+++ b/source/hic_hal/freescale/k20dx/gpio.c
@@ -1,6 +1,6 @@
 /**
  * @file    gpio.c
- * @brief   
+ * @brief
  *
  * DAPLink Interface Firmware
  * Copyright (c) 2009-2016, ARM Limited, All Rights Reserved
@@ -57,16 +57,12 @@ void gpio_init(void)
     PIN_nRESET_GPIO->PDDR &= ~PIN_nRESET;
     PIN_nRESET_PORT->PCR[PIN_nRESET_BIT] = PORT_PCR_MUX(1);
 
-    // Keep powered off in bootloader mode
-    // to prevent the target from effecting the state
-    // of the reset line / reset button
-    if (!daplink_is_bootloader()) {
-        // configure pin as GPIO
-        PIN_POWER_EN_PORT->PCR[PIN_POWER_EN_BIT] = PORT_PCR_MUX(1);
-        // force always on logic 1
-        PIN_POWER_EN_GPIO->PDOR |= 1UL << PIN_POWER_EN_BIT;
-        PIN_POWER_EN_GPIO->PDDR |= 1UL << PIN_POWER_EN_BIT;
-    }
+    // configure pin as GPIO
+    PIN_POWER_EN_PORT->PCR[PIN_POWER_EN_BIT] = PORT_PCR_MUX(1);
+    // set output to 0
+    PIN_POWER_EN_GPIO->PCOR = PIN_POWER_EN;
+    // switch gpio to output
+    PIN_POWER_EN_GPIO->PDDR |= PIN_POWER_EN;
 
     // Let the voltage rails stabilize.  This is especailly important
     // during software resets, since the target's 3.3v rail can take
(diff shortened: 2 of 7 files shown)


c1728p9:
@flit this change requires that FRDM boards are connected to a USB port with a data connection for the device to be powered right? That means you wouldn't be able to run off of a wall charger or portable battery pack. That seems like a fairly big downside to this change.

flit:
@c1728p9 That is a good point—you do actually need a USB host to configure the device. On the other hand, it's not USB compliant to (attempt to) draw more than 100 mA prior to configuration. For some PCs, boards that draw > 100 mA will simply not power up and are unusable. This is clearly not acceptable for customers, so we need some solution.

Some boards have a 5V power jack and/or can be powered via the target's USB connector. But not all.

Maybe we could have a timer? If USB is not configured within a certain time, then go ahead and flip the power switch.

flit:
@c1728p9 @cederom Do either of you have any preference about how to work around this? We definitely need to be USB compliant by default for high power boards, so must wait for USB configuration to enable the power switch.

Options:

1. Add macro so power switching after USB config is only done on boards where max current draw is greater than 100 mA.
2. Add timeout for USB config. After timeout, we enable the power switch anyway.
3. Combination of above two.
4. Other options?

Keep in mind that for boards that can accept Arduino shields or other expansion boards, you can easily exceed the 100 mA limit. So the first option may not be very good, as you could only apply it to boards that did not have expansion capabilities.

flit:
@c1728p9 @cederom Added 5 second timeout for USB config, after which target power is enabled.

cederom:
Hello @flit :-)

The first thing that comes to my mind is some sort of board-specific fix/quirk in board init/setup if the problem relates only to that particular board.. not necessairly to all boards out there..

If there are two USB connectors on the board, at least one of them should be used to power the board.. even if USB is not configured.. that may be user design that does not use USB at all but still wants to take power from USB port and power adapter as @c1728p9 mentioned..

From what  I understand the problem is with some target powered from DAPLink USB connector takes too much of energy.. but when it is connected with DAPLink and interface firmware works then what is exactly the problem? Could you please explain in detail? Is it about Target draining more than 100mA with no USB configured? If so this looks like target firmware or board issue..?

flit:
@cederom In short, the problem was that DAPLink on the K20 HIC was not USB compliant. The interface app would always turn on the target power switch prior to enabling USB. If the board drew more than 100 mA between when the switch was enabled and when USB was configured, certain PCs would disable the port.

We noticed this problem on the MIMXRT1050-EVK board since it is particularly power hungry. However, any board that has expansion capabilities (i.e., Arduino shield) can easily draw more than 100 mA. So this is a problem that most users could face.

The 5 second timer is to support powering via a USB wall wart—specifically the case that @c1728p9 mentioned. If DAPLink is running on the K20 HIC, then by design we know USB has been connected. If USB is not configured within 5 seconds, it's almost certainly because there is not a USB host.

If you can suggest a better way to identify that a USB power source without a host has been connected, I'll be happy to hear it! Maybe we could disable the timer once USB communication has started (first SETUP packet received from host), rather than when USB is configured. The downside to this is that it would require modifications to the USB stack instead of being entirely contained in main.c.

Fyi, we're going to production with this version of the firmware due to time constraints. (The board support is on commit flit/DAPLink@9cf8778c8f8f086, which will eventually be upstreamed.)

cederom:
Thanks for more details @flit :-) Why don't we just power on the target right after USB is configured.. and keep the target in reset or power down state until then? :-)

flit:
Hi @cederom, your suggestion to keep the target in reset or power down until USB is configured, at which point we power on the target, is exactly what this PR originally did. (And still does, with the addition of the timeout.) But without the USB configuration timeout, we can't support USB wall warts.

cederom:
ah, so this is the timeout not delay, then all fine..? :-)

flit:
Right 😄  It's just a timeout, not a delay.

flit:
Hi @c1728p9, Do you have any concerns about the implementation? Have you had a chance to review the code?

source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -57,24 +57,43 @@ void gpio_init(void)
...
-        PIN_POWER_EN_GPIO->PDOR |= 1UL << PIN_POWER_EN_BIT;
-        PIN_POWER_EN_GPIO->PDDR |= 1UL << PIN_POWER_EN_BIT;
+        // set output to 0
+        PIN_POWER_EN_GPIO->PCOR = PIN_POWER_EN;
+        // switch gpio to output
+        PIN_POWER_EN_GPIO->PDDR |= PIN_POWER_EN;
     }
+}
+
+void gpio_handle_usb_connected(gpio_usb_connect_state_t state)
```
c1728p9: Does the gpio driver need to know if USB is connected or if the USB connection timed out? If not, what about making this a function just for enabling and disabling board power, something like ```gpio_set_board_pwr()```? That would also make the code in main.c more explicit that it is enabling board power.

source/daplink/interface/main.c
```diff
@@ -57,13 +57,17 @@
...
 #define FLAGS_LED_BLINK_30MS    (1 << 6)
+
 // Timing constants (in 90mS ticks)
-// USB busy time
+// USB busy time (~3 sec)
 #define USB_BUSY_TIME           (33)
-// Delay before a USB device connect may occur
+// Delay before a USB device connect may occur (~1 sec)
 #define USB_CONNECT_DELAY       (11)
+// Timeout for USB being configured (~5 sec)
```
c1728p9: Any reason 5 seconds was chosen? This seems like a fairly long time. Could this delay be made shorter? Would ~1s be too short?

flit: 5 seconds just sounded reasonable at the time between me and Melissa Hunter (apps engineer). :) A longer timeout gives a pretty good indication that there really isn't a USB host, which was my primary concern. But I can certainly make it shorter. Maybe 2 or 2.5 seconds? That seems short enough that a user wouldn't notice.

source/hic_hal/freescale/k20dx/gpio.c
flit: I did it this way in case there were other use cases for the HIC to know the USB connection state.  I can change it if you prefer.

source/daplink/interface/main.c
c1728p9: 2 or 2.5 seconds sounds good. I'll try it out this week to get a feel for how noticeable this is.

flit: Ok, thanks.

source/hic_hal/freescale/k20dx/gpio.c
c1728p9: Did you have any other potential use cases in mind? If not it might be best to limit the scope to just enabling power.

flit: Not at the moment, no. I'll go ahead and change it. Thanks for the feedback.

flit:
@c1728p9 Do you think this change should be applied to the KL26Z HIC? I didn't want to touch that since it affects the microbit's behaviour. But on second thought, users can power external devices via the microbit's expansion header, so the same issue may be encountered.

flit:
Changed timeout to 2 seconds and changed gpio API to `gpio_set_board_power(bool powerEnabled)`. Rebased and squashed.

source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -57,24 +57,36 @@ void gpio_init(void)
...
+            // enable power switch
+            PIN_POWER_EN_GPIO->PSOR = PIN_POWER_EN;
+
+            // Let the voltage rails stabilize.  This is especailly important
+            // during software resets, since the target's 3.3v rail can take
+            // 20-50ms to drain.  During this time the target could be driving
+            // the reset pin low, causing the bootloader to think the reset
+            // button is pressed.
+            // Note: With optimization set to -O2 the value 1000000 delays for ~85ms
+            busy_wait(1000000);
```
c1728p9: This delay is needed specifically for the bootloader during bootloader init. Is it also needed when enabling power?

source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -57,24 +57,36 @@ void gpio_init(void)
...
-    // Let the voltage rails stabilize.  This is especailly important
-    // during software resets, since the target's 3.3v rail can take
-    // 20-50ms to drain.  During this time the target could be driving
-    // the reset pin low, causing the bootloader to think the reset
-    // button is pressed.
-    // Note: With optimization set to -O2 the value 1000000 delays for ~85ms
-    busy_wait(1000000);
+void gpio_set_board_power(bool powerEnabled)
+{
+    if (!daplink_is_bootloader()) {
```
c1728p9: Why is this turned off for bootloaders? Seems like it applies equally to bootloader an interface, just the interface code is the one which will turn it on.

source/hic_hal/freescale/k20dx/gpio.c
```diff
@@ -57,24 +57,36 @@ void gpio_init(void)
     PIN_nRESET_GPIO->PDDR &= ~PIN_nRESET;
     PIN_nRESET_PORT->PCR[PIN_nRESET_BIT] = PORT_PCR_MUX(1);
 
-    // Keep powered off in bootloader mode
-    // to prevent the target from effecting the state
-    // of the reset line / reset button
     if (!daplink_is_bootloader()) {
```
c1728p9: Since the code in this conditional isn't turning on power it doesn't need to be disabled for the bootloader.

source/hic_hal/freescale/k20dx/gpio.c
flit: Ok, my misunderstanding. I thought the delay was for waiting for the voltage rails to stabilize _after enabling power_. So it's probably not really necessary for enabling power. I'll move it up to  `gpio_init()` as before.

source/hic_hal/freescale/k20dx/gpio.c
flit: Partially a carry-over from the original code. Also, "board power" may mean different things to different boards. On some boards, I can imagine you would need the bootloader to turn on some power. So the bootloader might call this function just like the interface, and let the board decide whether to turn on power in the bootloader.

But I'll remove the bootloader check. Right now the bootloader doesn't call this function, and there are no boards that require additional power for the bootloader.

source/hic_hal/freescale/k20dx/gpio.c
flit: Ok, will change.

Generally I was trying to keep things as close to the original as possible to reduce risk. But I'm happy to clean it up if you're comfortable. (Of course, you have the board farm, so that helps ;)

flit:
Applied changes from review commands, rebased, and squashed.

PR #330, opened 2017-10-18
Update to CMSIS-DAP source from CMSIS 5

flit:
This patch synchronizes the standard CMSIS-DAP sources in source/daplink/cmsis-dap with the version of CMSIS-DAP from CMSIS 5. New features are the DAP_ExecuteCommands command and SWO support. Note that SWO is not actually supported by any HICs yet.


cederom:
@flit @c1728p9 I am working on SWD lock mechanism in DAPLink to prevent concurrent access. I would like to commit this change in order to put my changes on top.

At the moment I cannot see sensible way to implement lock mechanism without modification of this part of the code, which is used in HID access to Debug. DAPLink also uses SWD in UMS access to Flash firmware, and CDC+Reset handler in order to Reset target..

I would like to perform local changes to the DAPLink source tree to achieve my goal. Then propose upstream changes to CMSIS-DAP. Then merge those changes already containing lock mechanism back here when ready. Is that okay?

flit:
@cederom Well, I guess I don't understand the benefit of modifying this PR to include the lock support, instead of separately merging this PR followed by a PR adding the lock support.

cederom:
yea @flit ! that is exactly what i mean, tho have this pr first included! :-) :-)

c1728p9:
Thanks for making this update @flit it looks good!

PR #255, opened 2017-03-29
Lpc flash support

c1728p9:
Add flash support to the lpc11u35 interface so it can store settings.

PR #334, opened 2017-10-30
Disable interrupts when performing flash ops

c1728p9:
Update the atsam3u and kinetis flash drivers so they disable interrupts when erasing or programming flash. This prevents a crash from occurring if an interrupt occurs during flashing when the flash cannot be read.

This also fixes a watchdog reset on K20 interfaces which occurs when an interrupt interrupts the watchdog unlock sequence in the flash "Init" function.  The addition of this critical section prevents the watchdog operation from being interrupted.


c1728p9:
@mmahadevan108 in the process of making this change I noticed that the lpc4322 is also missing this lock and other updates.  You may want to implement this in another PR. For reference this change was made for the lpc11u35 in #255.

PR #343, opened 2017-11-04
Add MIMXRT1050-EVK board support

flit:
Note that there are two builds for the same board, reflecting the two memory population options of either QuadSPI or HyperFlash.


source/target/freescale/target_reset_mimxrt.c
```diff
@@ -0,0 +1,54 @@
...
+
+
+void target_before_init_debug(void)
+{
+    swd_set_target_reset(0);
+}
+
+void prerun_target_config(void)
+{
+    target_set_state(RESET_RUN);
```
c1728p9: Why does the target need to be put into the running state on boot? Is it not running already?

flit: Good question. I didn't do the original port, so I'll ask the author (Alex Yang). Will get back to you.

c1728p9: @flit any updates on this?

flit:
- Rebased.
- Moved `target_set_state(RESET_RUN)` to end of `target_before_init_debug()`.
- Added some comments (from Alex Yang) describing the purpose of the calls in `target_before_init_debug()`.

flit:
All CI tests passed, but results failed to be reported back to GitHub for some reason.

PR #344, opened 2017-11-08
Prevents concurrent access to the JTAG/SWD DAP.

cederom:
This solves concurrent access to Target's DAP over SWD (possibly also JTAG) from functions like HID Debug / CDC Reset / UMS Flashing.. with a minimal possible set of code and changes I could work out. Minor impact on CMSIS-DAP is presented with an example of application, when accepted can be extended and merged into upstream project..

Locking DAP prevents concurrent thread access to SWD/JTAG DAP operations. Lock may be assigned to Task (prevents concurrency) and to mark Operation. Task Lock is tied to OS and has precedence over user set Operation Lock. All underlying DAP/SWD/JTAG operations should be protected internally with provided lock verify mechanisms (TID based) that tells if DAP is available. By setting Operation Lock user also locks DAP for a given TID. No other TID can access DAP in that time, nor other Operation Lock can be set or clear, until ongoing Operation is unlocked from an originating owner thread/task. Unlocking the Operation makes DAP again available for use by anyone. Functions return 1 for OK/GO/READY, 0 for ERROR/BUSY/FAIL. User needs to initialize mutex by hand before any function can be called.

To demonstrate that lock is working, when accessing DAP that is busy you should get an assert from dap_lock.c or FAIL.TXT error with "Requested Resource Busy" (i.e. when reset or flashing occurs during pyOCD session).. this may be removed if not really necessary to inform user that way but comes handy for solution assessment..

Feedback testing and comments are welcome :-)

Signed-off-by: CeDeROM Tomasz CEDRO <tomek@cedro.info>

### DIFF
--- a/source/daplink/cmsis-dap/DAP.c
+++ b/source/daplink/cmsis-dap/DAP.c
@@ -34,6 +34,7 @@
 #endif
 #include "DAP_config.h"
 #include "DAP.h"
+#include "swd_host.h"
 
 
 #define DAP_FW_VER      "1.10"  // Firmware Version
@@ -82,11 +83,33 @@ const char TargetDeviceName   [] = TARGET_DEVICE_NAME;
 #endif
 
 
+//////////////////////////////////////////////////////////////////////////////
+// DAP Lock routines for use in mutitasking environment.
+// Prevents concurrent access to DAP Port like Mutex.
+// Marked __weak to for easy implementation specific override.
+// Returns 1 for OK, 0 for Error
+
+__weak uint8_t DAP_Lock_Verify(void *param) {
+	return 1;
+}
+
+__weak uint8_t DAP_Lock(void *param) {
+	return 1;
+}
+
+__weak uint8_t DAP_Unlock(void *param) {
+	return 1;
+}
+
+// DAP Lock section ends
+//////////////////////////////////////////////////////////////////////////////
+
 // Get DAP Information
 //   id:      info identifier
 //   info:    pointer to info data
 //   return:  number of bytes in info data
 static uint8_t DAP_Info(uint8_t id, uint8_t *info) {
+	if (!DAP_Lock_Verify(0)) return 0;
   uint8_t length = 0U;
 
   switch (id) {
@@ -168,6 +191,8 @@ static uint32_t TimerTick;
 
 // Start Timer
 static __inline void TIMER_START (uint32_t usec) {
+	if (!DAP_Lock_Verify(0)) return 0;
+
   TimerTick = osKernelSysTick() + osKernelSysTickMicroSec(usec);
 }
 
@@ -176,6 +201,8 @@ static __inline void TIMER_STOP (void) {}
 
 // Check if Timer expired
 static __inline uint32_t TIMER_EXPIRED (void) {
+	if (!DAP_Lock_Verify(0)) return 0;
+
   return ((osKernelSysTick() > TimerTick) ? 1U : 0U);
 }
 
@@ -183,6 +210,8 @@ static __inline uint32_t TIMER_EXPIRED (void) {
 
 // Start Timer
 static __inline void TIMER_START (uint32_t usec) {
+	if (!DAP_Lock_Verify(0)) return;
+
   SysTick->VAL  = 0U;
   SysTick->LOAD = usec * (CPU_CLOCK/1000000U);
   SysTick->CTRL = (1U << SysTick_CTRL_ENABLE_Pos) |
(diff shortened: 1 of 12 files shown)


source/usb/cdc/usbd_core_cdc.c
```diff
@@ -93,9 +94,18 @@ __weak BOOL USBD_EndPoint0_Setup_CDC_ReqToIF(void)
                 break;
 
             case CDC_SEND_BREAK:
+								if (!dap_lock_operation(DAP_LOCK_OPERATION_CDC_BREAK))
+								{
+									util_assert(0);
```
c1728p9: Asserts should only trigger when a DAPLink bug occurs. This is asserting on user error - break is sent while something else is using SWD.

source/daplink/cmsis-dap/DAP.c
```diff
@@ -82,11 +83,33 @@ const char TargetDeviceName   [] = TARGET_DEVICE_NAME;
...
+
+// DAP Lock section ends
+//////////////////////////////////////////////////////////////////////////////
+
 // Get DAP Information
 //   id:      info identifier
 //   info:    pointer to info data
 //   return:  number of bytes in info data
 static uint8_t DAP_Info(uint8_t id, uint8_t *info) {
+	if (!DAP_Lock_Verify(0)) return 0;
```
c1728p9: 4 spaces should be used in place of a tab. Same comment applies to code below.

source/daplink/cmsis-dap/DAP.c
```diff
@@ -168,6 +191,8 @@ static uint32_t TimerTick;
 
 // Start Timer
 static __inline void TIMER_START (uint32_t usec) {
+	if (!DAP_Lock_Verify(0)) return 0;
```
c1728p9: This is application specific code that probably shouldn't be in this file. If you want to add checking I would leave it as a sanity check assert which doesn't effect program flow. This shouldn't be hit in practice if locking is done correctly in the layer above.

source/daplink/interface/main.c
```diff
@@ -213,6 +213,9 @@ __task void main_task(void)
     uint8_t thread_started = 0;
     // button state
     main_reset_state_t main_reset_button_state = MAIN_RESET_RELEASED;
+    // Initialize DAP locks.
+    util_assert(dap_lock_mutex_init());
```
c1728p9: Functions with side effects should be kept out of asserts, so when asserts are turned off the program behavior does not change.

source/daplink/interface/swd_host.c
```diff
@@ -117,13 +123,26 @@ uint8_t swd_init(void)
 
 uint8_t swd_off(void)
 {
-    PORT_OFF();
+	// Unlock SWD Port by current TID, if possible.
```
c1728p9: old comment

source/daplink/interface/swd_host.c
```diff
@@ -145,6 +164,13 @@ uint8_t swd_read_dp(uint8_t adr, uint32_t *val)
 // Write debug port register
 uint8_t swd_write_dp(uint8_t adr, uint32_t val)
 {
+	// Verify DAP Lock against our TID.
+	if (!dap_lock_verify_tid_self())
```
c1728p9: These should be passive asserts as well.

source/daplink/cmsis-dap/DAP.c
```diff
@@ -82,11 +83,33 @@ const char TargetDeviceName   [] = TARGET_DEVICE_NAME;
 #endif
 
 
+//////////////////////////////////////////////////////////////////////////////
+// DAP Lock routines for use in mutitasking environment.
+// Prevents concurrent access to DAP Port like Mutex.
+// Marked __weak to for easy implementation specific override.
+// Returns 1 for OK, 0 for Error
+
+__weak uint8_t DAP_Lock_Verify(void *param) {
```
c1728p9: What is param for?

source/daplink/cmsis-dap/DAP.c
```diff
@@ -1467,10 +1547,12 @@ uint32_t DAP_ProcessCommand(const uint8_t *request, uint8_t *response) {
       break;
 
     case ID_DAP_Connect:
+			if (!DAP_Lock(0)) return 0;
```
c1728p9: If the disconnect message is missed then DAP will remain locked. There may need to be some mechanism to timeout or to steal ownership of the lock. A similar problem also applies CDC.

brianesquilona:
Will close this PR due to incomplete implementation, though feel free to reopen it and rebase.

PR #351, opened 2017-11-21
Simplify reset button handling

c1728p9:
Rather than handling forwarded reset in on vendor specific code, instead handle this at a common layer. To do this, this patch adds the functions gpio_get_reset_btn_no_fwrd() and gpio_get_reset_btn_fwrd() which are far easier to implement.


c1728p9:
Rebased and added gpio_get_reset_btn() which is true if either gpio_get_reset_btn_no_fwrd() or gpio_get_reset_btn_fwrd() is true.

c1728p9:
Tested locally and verified the reset button works correctly on the lpc11u35, lpc4322, microbit, k20 and atsam3u interfaces.

cederom:
good change! :-)

PR #353, opened 2017-11-22
Add the ncs36510 to the released firmware list

c1728p9:
Add the ncs36510 firmware to the public release package.


c1728p9:
CC @studavekar

Issue #340, opened 2017-11-02
LPC4322 IAP functions not protected


c1728p9:
The lpc4322 is missing a critical section around IAP calls. Additionally the memory map does not block off ram for IAP calls. For reference these changes were made for the lpc11u35 in #255.

c1728p9:
@mmahadevan108 can you make these updates since you are maintaining the lpc4322?

c1728p9:
fixed by #358

PR #358, opened 2017-11-30
LPC4322 Protect IAP function calls

mmahadevan108:
Fix for Issue 340

Signed-off-by: Mahesh Mahadevan <Mahesh.Mahadevan@nxp.com>

### DIFF
--- a/source/hic_hal/nxp/lpc4322/flash.c
+++ b/source/hic_hal/nxp/lpc4322/flash.c
@@ -22,6 +22,7 @@
 #include "flash_hal.h"
 #include "system_LPC43xx.h"
 #include "daplink_addr.h"
+#include "cortex_m.h"
 
 #define END_SECTOR     14  /* 15 sectors per bank */
 #define FLASH_BANK_A    0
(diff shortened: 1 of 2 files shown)


mmahadevan108:
This is a fix for https://github.com/ARMmbed/DAPLink/issues/340

c1728p9:
Thanks for the PR @mmahadevan108. I think the function ```read_unique_id``` also uses IAP calls and also needs to be protected. Do any of the IAP calls use global ram which needs to be reserved? This was needed for the lpc11u35 - https://github.com/ARMmbed/DAPLink/pull/255/files#diff-8d605fbb650e7188d924bb38e686e453R31.

c1728p9:
@mmahadevan108 any updates on this?

mmahadevan108:
@c1728p9 I have updated the patch to protect the IAP calls in read_unique_id function. We do not reserve any global ram for lpc4322 driver.

c1728p9:
Hi @mmahadevan108, looking at the LPC4322 datasheet it looks like RAM does need to be reserved.

From sections 6.8 of UM10503:
**IAP commands which result in a flash write/erase operation use 32 bytes of space in the top portion of the on-chip RAM for execution. The user program should not be using this space if IAP flash programming is permitted in the application.**

mmahadevan108:
The flash programming code uses RAM located at 0x10080000. This SRAM is not used by LPC4322 DAPLink code.

c1728p9:
Ah good, I didn't realize this was in a separate RAM bank.

mmahadevan108:
Can this be merged in?

PR #360, opened 2017-12-04
Update readme content.

sg-:
Fix broken links, update text and add a fancy image. A view of this can be seen on my branch https://github.com/sg-/DAPLink/tree/update-readme

Replacement for #356


c1728p9:
@sg- Is there any way a sharper image can be used? It looks kinda fuzzy. Also, it may be good to have a line or something separating it from the text below, but more a matter of preference.

sg-:
Working on a sharper image now

PR #361, opened 2017-12-06
ff_lpc546xx: add new target

gorazdko:
add new target L-Tek FF-LPC546XX
board id: 8081


c1728p9:
Hi @gorazdko can you attach test logs for the new boards you are adding?

PR #362, opened 2017-12-08
Keep interrupt priority group value unchanged when writing to the AIRCR register

mmahadevan108:
The issue was uncovered when issuing a SYSTEM RESET on the i.MXRT platform.

Signed-off-by: Mahesh Mahadevan <Mahesh.Mahadevan@nxp.com>

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -52,6 +52,10 @@
 #define MAX_TIMEOUT   1000000  // Timeout for syscalls on target
 
 #define SOFT_RESET  SYSRESETREQ
+
+#define SCB_AIRCR_PRIGROUP_Pos              8                                             /*!< SCB AIRCR: PRIGROUP Position */
+#define SCB_AIRCR_PRIGROUP_Msk             (7UL << SCB_AIRCR_PRIGROUP_Pos)                /*!< SCB AIRCR: PRIGROUP Mask */
+
 // Some targets require a soft reset for flash programming (RESET_PROGRAM).
 // DAP operations as they are controlled by the remote debugger.
 #if defined(BOARD_BAMBINO_210) || defined(BOARD_BAMBINO_210E) || defined(TARGET_NRF51822)
@@ -1005,7 +1009,11 @@ uint8_t swd_set_target_state_sw(TARGET_RESET_STATE state)
             }
 
             // Perform a soft reset
-            if (!swd_write_word(NVIC_AIRCR, VECTKEY | SOFT_RESET)) {
+            if (!swd_read_word(NVIC_AIRCR, &val)) {
+                return 0;
+            }
+
+            if (!swd_write_word(NVIC_AIRCR, VECTKEY | (val & SCB_AIRCR_PRIGROUP_Msk) | SOFT_RESET)) {
                 return 0;
             }
 


mmahadevan108:
cc @c1728p9 @flit

flit:
What was the symptom of this bug? I'm actually surprised `PRIGROUP` doesn't get reset automaticaly by the core when you perform the system reset.

mmahadevan108:
We would see a FAIL.TXT file with the message "The interface firmware FAILED to reset/halt the target MCU"

flit:
Do you know what the root cause was? Is it simply that you can't change any other fields of AIRCR when setting `VECTRESET`?

mmahadevan108:
The failure to reset/halt the target MCU is seen when using SYSTEMRESET without the patch to retain interrupt priority bits.

c1728p9:
The priority bits should get reset when a software reset occurs. This is from the ARMv7m architecture reference manual:
```
PRIGROUP, bits[10:8]
Priority grouping, indicates the binary point position.
For information about the use of this field see Priority grouping on page B1-583.
This field resets to 0b000.
```

Can you confirm with a debugger or logging code in DAPLink that these bits aren't getting reset?

flit:
The change is not because `PRIGROUP` isn't getting reset, it is to perform a read-modify-write so that only `VECTRESET` is changed in the write to AIRCR. After the core resets, `PRIGROUP` should read as zeros.

I think the issue here is that you are not supposed to modify any other writable bits in the AIRCR when you are setting `VECTRESET` or the behaviour is undefined. The v7-M ARM is not very clear about this, but it makes sense given the way some of the other registers work. Also, this may be different on the CM7 versus other cores.

c1728p9:
@flit this patch only preserves the bits in SCB_AIRCR_PRIGROUP_Msk, the others are set to 0 as they were before. Because of this, I'm a little surprised this fixes the issue.

mmahadevan108:
The CMSIS code for System Reset has a similar implementation:
https://github.com/ARMmbed/mbed-os/blob/master/cmsis/TARGET_CORTEX_M/core_cm7.h

/**
  \brief   System Reset
  \details Initiates a system reset request to reset the MCU.
 */
__STATIC_INLINE void __NVIC_SystemReset(void)
{
  __DSB();                                                          /* Ensure all outstanding memory accesses included
                                                                       buffered write are completed before reset */
  SCB->AIRCR  = (uint32_t)((0x5FAUL << SCB_AIRCR_VECTKEY_Pos)    |
                           (SCB->AIRCR & SCB_AIRCR_PRIGROUP_Msk) |
                            SCB_AIRCR_SYSRESETREQ_Msk    );         /* Keep priority group unchanged */
  __DSB();                                                          /* Ensure completion of memory access */

  for(;;)                                                           /* wait until reset */
  {
    __NOP();
  }
}

mmahadevan108:
cc @maclobdell

mmahadevan108:
Can this be merged?

c1728p9:
Hi @mmahadevan108 this fix needs to be understood and verified before it can be merged. Can you confirm with a debugger or logging code in DAPLink that these bits aren't getting reset? Alternatively, if you can provide me with the steps to reproduce this then I can try and replicate this locally. If it can be confirmed that this behavior in not to spec then I can escalate this issue.

mmahadevan108:
The reset code implementation was changed to match what is present in the CMSIS code. Could you help check with the team in ARM that provides this.

c1728p9:
I opened an issue in the CMSIS_5 repo to find out more information. In the mean time could you either confirm this locally or provide me with steps to reproduce the issue?

flit:
@c1728p9 Can this PR be merged so the reset code is aligned with CMSIS_5?

To respond to an earlier comment from you, PRIGROUP is the only field of AIRCR that is r/w. Thus it is the only field that would need to be preserved in a read-modify-write.

c1728p9:
Hi @mmahadevan108 this isn't a failure due to this PR. I'm working to get CI fixed.

c1728p9:
@mmahadevan108 this time the build failure is legitimate. If you try to build kl26z_microbit_if locally you should see these build failures.

mmahadevan108:
Thank you, the priority group bits do not exist on Cortex-M0 & Cortex-M0+. I have added a check to account for this.

source/daplink/interface/swd_host.c
```diff
@@ -1005,10 +1005,19 @@ uint8_t swd_set_target_state_sw(TARGET_RESET_STATE state)
             }
 
             // Perform a soft reset
-            if (!swd_write_word(NVIC_AIRCR, VECTKEY | SOFT_RESET)) {
+#if (__CORTEX_M >= 0x03)
```
c1728p9: This define is for the MCU running DAPLink when really it depends on the target MCU. Could you define ```SCB_AIRCR_PRIGROUP_Msk``` at the top of this file so it is present for all MCUs?

c1728p9:
@mmahadevan108 could you update this PR again?

PR #363, opened 2017-12-11
Add Web USB support to DAPLINK

arekzaluski:
- Merge with current master
- Add new Web USB HID endpoint

demo available on:
https://armmbed.github.io/dapjs-web-demo/

Tested operating systems:
Windows 8.1, 10
Mac OS X 10.10, 10.12
Ubuntu 14.04, 16.04
Tested platforms:
FRDM-K64F, Microbit

Known issues:

- Linux: usb devices are read only by default, permission : 0644.  Udev rule with permission 0666 required. See more information here: https://developers.google.com/web/updates/2016/03/access-usb-devices-on-the-web
- Windows: Landing page is not working. More information here: https://github.com/WICG/webusb/issues/104

@sg- @c1728p9 @cederom @thegecko


cederom:
Hey @arekzaluski we are looking into changes soon as we are together with @c1728p9 in AU/TX :-)

I can see there are some conflicts you probably first need to fix.. why there are so many changes/commits? Does it change the codebase a lot or simply adds something aside next to what we have?

flit:
Looks like most of those commits are coming from the merge from master.

@arekzaluski You should rebase on top of master instead of merge. This will reduce the commits to only those for your changes and make your PR a lot easier to review. After you rebase locally, do a force push to GitHub to update this PR.

arekzaluski:
@cederom @flit Yes, I wanted to keep the history of commits.
It's still a proof of concept. We will need your work to make it a fully working solution. That PR is to armmbed_webusb branch. Not to master.

flit:
On a different question, I'm curious why a new HID endpoint is required for WebUSB? Can WebUSB not access the standard CMSIS-DAP HID endpoint for some reason?

Also note that many host USB libraries have issues with devices having multiple HID interfaces. For instance, pywinusb crashes, and hidapi on macOS may have issues (but hidapi on Windows works). I've had to deal with these issues with the default NXP LPC-Link2 CMSIS-DAP firmware that has 3x HID interfaces.

In addition, devices with multiple HID interfaces fail USB compliance testing. (I heard this anecdotally regarding the LPC-Link2 firmware.)

arekzaluski:
Default HID interface (3) is claimed by OS. In WebUSB you can't release it and claim for yourself. That's why in this solution we have a new interface (255) that is connected to CMSIS-DAP. They work alongside each other. There is boolean flag that sends the data to the correct interface.
Similarly later we want to add also another interface for CDC.

flit:
Ok, got it. It's sort of like an endpoint alias.

Please make sure to set the HID usage page for the WebUSB HID interface to something unique other than 0xFF00. This will allow us to properly identify the regular HID interface versus WebUSB. (USB interfaces can also have names, but you can't get those names via hidapi.)

Fyi, the [dev_lpc_link](https://github.com/flit/pyOCD/tree/dev_lpc_link) branch in my pyOCD fork adds a filter so it only sees HID interfaces with a usage page of 0xFF00. This was also necessary for connecting to the LPC-Link2 CMSIS-DAP firmware.

arekzaluski:
@flit If it helps we can use also different interface number. 0xFF is designated for vendor specific. We can use a undefined interface like 0xAA

flit:
I'm not sure if you can get the interface number from hidapi, will have to check. But I'm hesitant to use the interface number since that could theoretically change. We could just use a usage page of 0xFF01 (vendor specific page 1) or something similar for the WebUSB interface.

thegecko:
👍

PR #368, opened 2017-12-13
Add 96boards Nitrogen, board id 0x9014

rsalveti:
Board details available at https://www.96boards.org/product/nitrogen

Signed-off-by: Ricardo Salveti <ricardo@opensourcefoundries.com>

### DIFF
--- a/records/board/96b_nitrogen.yaml
+++ b/records/board/96b_nitrogen.yaml
@@ -1,0 +1,11 @@
+common:
+    macros:
+        - IO_CONFIG_OVERRIDE
+    includes:
+        - source/board/override_96b_nitrogen
+    sources:
+        board:
+            - source/board/96b_nitrogen.c
+        target:
+            - source/target/nordic/nrf5x/target.c
+            - source/target/nordic/target_reset_nrf52.c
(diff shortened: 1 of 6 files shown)


records/board/96b_nitrogen.yaml
```diff
@@ -0,0 +1,11 @@
+common:
+    macros:
+        - IO_CONFIG_OVERRIDE
```
c1728p9: What IO customization did you need to make for this board? Please comment in IO_Config_Override.h which pins changed from the default.

test/info.py
```diff
@@ -140,6 +140,7 @@
     (   0x9011,     'lpc11u35_archmax_if',                      None,               'Seeed-Arch-Max'                        ),
     (   0x9012,     'lpc11u35_tiny_if',                         None,               'Seeed-Tiny-BLE'                        ),
     (   0x9013,     'lpc11u35_archlink_if',                     None,               'Seeed-Arch-Link'                       ),
+    (   0x9014,     'lpc11u35_96b_nitrogen_if',                 None,               'Seeed-96Boards-Nitrogen'               ),
```
c1728p9: If you want this as part of the next DAPLink release you'll need to add it to the above PROJECT_RELEASE_INFO table as well.

rsalveti: Added to the PROJECT_RELEASE_INFO table as suggested, thanks!

records/board/96b_nitrogen.yaml
rsalveti: Added "Nitrogen override" for the pins that changed from the default.

c1728p9:
Thanks for the updates @rsalveti

PR #372, opened 2017-12-15
Fix testing on Windows 10

c1728p9:
When storage service is turned on and System Volume Information is being created running chkdsk can cause drive caching to enter a bad state. This patch adds a workaround for this by re-running chkdsk until the drive is in a good state again.


cederom:
Hey @c1728p9 thanks for this update! Here goes the result of first run.. it seems to fail on one test.. I will retry and send updates :-)

```
One or more tests have failed

Git info for test:
  Git SHA: d6212b4916d98fecfe81c7da03a519c606b4bd3f
  Local changes: False

Test settings:
  Load application before test: True
  Load bootloader before test: False
  Run DAPLink specific tests: False
  Run endpoint tests: True

Tested configurations:
  APP=k20dx_frdmk64f_if BL=k20dx_bl Board=FRDM-K64F Target=FRDM-K64F: Fail

All firmware in package tested
```

[k20dx_frdmk64f_if FRDM-K64F.txt](https://github.com/ARMmbed/DAPLink/files/1564003/k20dx_frdmk64f_if.FRDM-K64F.txt)

cederom:
The problem seems to persist on my Win10 PC (not VM). Even after stopping StorageServices..

cederom:
With StorageServices disabled:
```
Test: k20dx_frdmk64f_if FRDM-K64F: Failure
  Test: test_endpoints: Failure
    Test: test_mass_storage: Failure
      Test: Load binary with flushes: Failure
        Failure: Device reported failure: "The transfer timed out."
      Test: Load hex with flushes: Failure
        Failure: Device reported failure: "The transfer timed out."
```
[k20dx_frdmk64f_if FRDM-K64F.txt](https://github.com/ARMmbed/DAPLink/files/1564054/k20dx_frdmk64f_if.FRDM-K64F.txt)

c1728p9:
This doesn't fix transfers timing out or out of order transfers but it does fix the problems with chkdsk.

PR #385, opened 2018-01-31
Add support for Rapid-IoT & Hexiwear boards

mmahadevan108:


mmahadevan108:
cc @flit @maclobdell @c1728p9

source/board/hexiwear.c
```diff
@@ -0,0 +1,60 @@
...
+
+void prerun_board_config(void)
+{
+    // SIM peripheral   0x40047000
+    // address offset   0x    1024
+    uint32_t SDID = 0x40048024;
+    uint32_t sdid;
+
+    // get a hold of the target
+    target_set_state(RESET_PROGRAM);
```
c1728p9: A reset on boot should be avoided, as this could corrupt any filesystem operations the target is doing.

source/target/freescale/rapid_iot/flash_blob_k64.c
```diff
@@ -0,0 +1,90 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "flash_blob.h"
+
+static const uint32_t K64F_FLM[] = {
```
c1728p9: Why is an additional K64F flash algo added?

records/board/hexiwear.yaml
```diff
@@ -0,0 +1,7 @@
+common:
+    sources:
+        board:
+            - source/board/hexiwear.c
+        target:
+            - source/target/freescale/rapid_iot/target.c
```
c1728p9: Why is the rapid_iot target used for the hexiwear?

mmahadevan108: Rapid-IoT is the next generation of Hewiwear.

source/board/hexiwear.c
mmahadevan108: This is required to be able to read the target registers.

source/target/freescale/rapid_iot/flash_blob_k64.c
mmahadevan108: Hexiwear has K64F and KW40Z.
Rapid-IoT has K64F and KW41Z.

Depending on the switch settings, user can flash to either K64F or KW4x, Hence we read the SDID register from the target to load the appropriate flash algo.

source/board/hexiwear.c
teetak01: You should not bring back any filesystem corruption bugs with additional resets at startup.

mmahadevan108:
This is needed in order to read target registers to decide the flash algorithm to use. It is during init before any flash operations.

cederom:
Why not use that inside RESET_PROGRAM before flashing?

There was serious issue recently with K64F being reset on reset that for a moment started and interrupted a target running software that caused filesystem corruption..

mmahadevan108:
cc @c1728p9 @teetak01 @flit 

Updated this PR to remove the reset and set the target flash algo inside RESET_PROGRAM

cederom:
I mean `target_set_state()` is a generic call for every Target. We have also rewritten that recently (https://github.com/ARMmbed/DAPLink/pull/371) so it should not be touched inside or altered. It was weak before but will be generic for all targets..

Instead, for target autodetection we could use something that is target specific, maybe `void target_before_init_debug()` from [1]?

[1] https://github.com/ARMmbed/DAPLink/blob/master/source/target/freescale/target_reset_Kseries.c

mmahadevan108:
I have updated the PR to move the code out of target_set_state()

source/target/freescale/rapid_iot/target_reset.c
```diff
@@ -0,0 +1,187 @@
...
+    // address offset   0x    1054
+    uint32_t UUID_LOC = 0x40048054;
+    uint32_t uuid[4] = {0};
+    // SIM peripheral   0x40047000
+    // address offset   0x    1024
+    uint32_t SDID = 0x40048024;
+    uint32_t sdid;
+
+    // get a hold of the target
+    target_set_state(RESET_PROGRAM);
```
c1728p9: The target is getting reset on boot which could cause problems for persistent storage.

mmahadevan108: This is there in the current code for Kinetis devices and not something that was added as part of this change.

c1728p9: This was removed in #388. You'll need to pull in that same change here.

mmahadevan108:
I added the below comment to the wrong PR. 
@c1728p9 Any timeline when code to read target register can be added. This impacts support for this product.

mmahadevan108:
@c1728p9 @cederom @teetak01 @maclobdell . Till we have a way to read target registers can this PR be included. The target reset call is limited to Rapid-IoT & Hexiwear boards.

mmahadevan108:
@c1728p9 I have updated the implementation using #398

JanneKiiskila:
When can this be merged? 
This is now blocking Mbed Cloud Client enablement.

c1728p9:
Hi @JanneKiiskila I'll start testing now and if everything looks good I'll merge.

JanneKiiskila:
Super!

mmahadevan108:
@c1728p9 I have rebased and updated this PR.

JanneKiiskila:
When will this be included in a release?

c1728p9:
@JanneKiiskila tentatively around the time of the next mbed-os release

PR #388, opened 2018-02-01
Fix Kinetis K series reset on boot

c1728p9:
Remove the code which performs a reset on boot since this risks corrupting any nonvolatile storage on the target device. This code is for reading the target ID and is no longer used.

### DIFF
--- a/source/target/freescale/target_reset_Kseries.c
+++ b/source/target/freescale/target_reset_Kseries.c
@@ -33,24 +33,6 @@ void target_before_init_debug(void)
     swd_set_target_reset(1);
 }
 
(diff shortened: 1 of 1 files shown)


mmahadevan108:
How do you recommend we read the target register to get ID information which is used to decide target flash algorithm.

c1728p9:
@mmahadevan108 to get the ID info the target could be halted and resumed without resetting. Unfortunately the current DAPLink API does support this. This feature needs to be added to DAPLink.

mmahadevan108:
Any timeline when this can be added. This impacts support for this product.

PR #398, opened 2018-02-22
Allow target to be halted without being reset

c1728p9:
Make changes to allow targets to be halted without a reset occurring:
- Reset after programming to update state
- Remove reset in Kinetis debug init
- Add options to HALT and RUN,


c1728p9:
@maclobdell @mmahadevan108 this should unblock #385. Please test and provide feedback.

mmahadevan108:
Thank you @c1728p9 . This works for  #385

Issue #269, opened 2017-05-15
Using daplink with Android over USB


DavidWhaleMEF:
microbit.org-ticket: 1558

Users are struggling to get Android to detect the micro:bit as a removable drive over USB. Some information below. Is this something we would expect to work, do you know? Thanks!

I would feel much happier using USB from a tablet and have been running some tests with USB OTG. So far I haven't been able to talk to the micro:bit, presumably because there is no software support for OTG in the micro:bit. Would you consider making the micro:bit compatible with USB OTG?

platform: Samsung Tab 4

Using USB OTG File Manager by Top Photo Inc I can read and write files to a memory stick using an OTG adaptor cable. 

When I connect the micro:bit instead of the memory stick the file manger displays the device as an "mbed", but doesn't show any files. I've tried copying a compiled .hex file to the mbed but it doesn't transfer.

Original firmware was:
DAPLink Firmware - see https://mbed.com/daplink
Version: 0234
Build:   Oct 12 2015 14:53:22

Updated firmware to: 
# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 9900000031864e45006d100b0000002a0000000097969901
HIC ID: 97969901
Auto Reset: 1
Automation allowed: 0
Overflow detection: 0
Daplink Mode: Interface
Interface Version: 0243
Git SHA: b403a07e3696cee1e116d44cbdd64446e056ce38
Local Mods: 0
USB Interfaces: MSD, CDC, HID
Interface CRC: 0x14256f44
Remount count: 0

Now it responds as follows:

It's behaviour with the tablet over USB OTG is now slightly different in that the micro:bit is displayed as an ARM mass storage device rather than an mbed. However I still can't read or write files to it using USB OTG File Manager by Top Photo Inc .

Also the yellow LED only comes on briefly when I connect USB whereas previously the yellow LED was on continuously. This point is probably insignificant, but I thought I'd note it just in case it's relevant.

Just a thought - my understanding is that USB OTG only supports FAT32  and NTFS file formats. I don't know what format(s) the micro:bit supports, but could this be the problem?

I've reformatted my memory stick to FAT16 and can read and write files to it from the tablet, even though Android isn't supposed to support FAT16. Maybe the file Manger I'm using does support this format, although it doesn't say so in the spec.

I've been unable to reformat my memory stick to FAT12.

DavidWhaleMEF:
[...]

PR #401, opened 2018-02-24
VFS/VBR/FAT Boot Sector Signature + Bootstrap Code / Android USB OTG FIX

cederom:
**This PR fixes Android USB OTG mount problem**
When Disk Boot Signature is not `0xAA55` (note hexdump order `55 AA`) then Android cannot mount DAPLink and considers it a broken drive. This prevented mount on Android USB OTG due to internal `fsck_msdos` failure on check before mount. This signature shall remain `0xAA55` to indicate Sector 0 end as defined by standard!

**This PR adds new PC Bootstrap Code**
As Boot Signature was marked `0xAA55` some systems may consider it bootable. Because "accidental" boot from USB DAPLink could cause trouble and confuse some users, I have created a simple PC Bootloader that prints out warning message on the computer screen when booted from USB DAPLink:
```
PLEASE REMOVE THE ARM MBED DAPLINK USB DEVICE AND REBOOT THE SYSTEM..
```
The bootloader source code and compiler script plus payload generator is attached in VFS sources. Someone with experience in x86 bootloader design should verify the code please. Modifications should be trivial to follow.

**Big Thank You to Microbit Community for pointing out the https://github.com/ARMmbed/DAPLink/issues/269 problem and Google Support + Developers for Testing and pointing out the hints on Android internals! :-)**

https://issuetracker.google.com/issues/73130696


jaustin:
This is fantastic, thanks @cederom !
@c1728p9 we'd be keen to know whether this is likely to make it into 0247? Do we need additional testing due to these changes?

brianesquilona:
Unfortunately having this signature makes some PC think that it can boot from the DAPLink drive, although the prints are useful, in CI, we need to configure the PC not to boot on some drives which is a bit of trouble on restart.

cederom:
Hello @brianesquilona, I understand the problem, here are some remarks:
* `0xAA55` signature is required by some `msdos_fsck` to verify clean filesystem and attach the drive (i.e. Android or FreeBSD). Because there was no signature Android considered disk to be broken.
* `0xAA55` may be used to mark disk bootable.. but its also used to detect valid filesystem.
* We are using VBR not the MBR.

Fixes may include:
* Configure CI machines BIOS/SETUP correctly:
   * Disable USB boot at all. You still may use USB boot with F12 or F8 or ESC (depends on machine) Boot Menu triggered by hand when necessary.
   * Use internal SATA or selected external Hard-Drive only as the only boot medium. Disable all other boot devices. That would prevent accidental boot from unwanted devices.
   * Use UEFI boot. That will ignore BIOS/MBR boot devices. You would have to use UEFI aware OS on the CI machines.
* Search for a bit in VBR structures that marks disk bootable (something else than `0xAA55` signature.. I think there was one specific bit that could be used here but I don't remember exactly now sorry).
* Maybe the bootloader code may return error value or trigger boot from a next device.
* Treat VBR as MBR, describe disk as first and only partition, but NOT active. That would give us both valid disk signature and non-bootable partition.
* Implement MBR in addition to VBR to have most desirable and versatile solution.
* Forget about MBR and switch to GPT. That would break backwards-compatibility with older systems.

flit:
Fwiw, I completely agree that the 0xaa55 signature is required. It's not a valid filesystem with this signature missing. We need to find some other mechanism to prevent PCs from trying to boot from the DAPLink volume.

jaustin:
Hey @brianesquilona did you find a workaround to the booting issues described above, hence the merge? Given the number of schools that want to leave micro:bits plugged in and reboot machines, etc, I'd like to make sure we know about the implications of this one in advance :)

cederom:
Hey Johny, the simplest solution would be to setup BIOS settings as presented above. DAPLink will print a message on screen so you clearly know when machine boots off it.. maybe there is another way, when found it will go to upstream..

jaustin:
@cederom thanks for the quick response, unfortunately with more than 2M micro:bits in different kinds of school systems now it's hard for us to make recommendations about bios setup - even just having to install a driver, or application, is a significant barrier for using tech in schools, and one of the reasons DAPLink helps so much!

Likewise, even with the helpful message printed, I think people will potentially be very confused - to most teachers or students users, the phrase "ARM MBED DAPLINK" doesn't mean (yet! hopefully the point is they'll learn much more as they get into coding) so may not even help them identify which device to remove.

...so if this change is going upstream, and machines will still try to boot from any connected DAPLink device, it would be really good if we could make the change optional at build time?

brianesquilona:
@jaustin the signature is a standard to comply with FAT standard. I don't know about other PC machines but on some Dell systems, without this signature, the drive was mistaken to be a diskette drive and on legacy boot, some machines will still boot from it even without the signature and will just print a general boot error on the screen. At least this PR prints an error if the system mistakenly booted on the DAPLink drive.

cederom:
This also makes DAPLink work on Android ;-)

flit:
We could potentially make the message configurable per board.

Without this change, the FAT filesystem from DAPLink is invalid and corrupt. I'm really surprised that any OSes will mount the volume. Android is doing the correct thing by reporting a corrupted filesystem and refusing to mount it.

We will look at adding an MBR and see if there's anything we can do there.

PR #406, opened 2018-03-12
Add WebUSB support

arekzaluski:
Add Microsoft OS 2.0 Descriptor support

* Update webusb landing page to os.mbed.com

@c1728p9 @cederom @sg- @flit @thegecko

### DIFF
--- a/source/usb/usbd_core.c
+++ b/source/usb/usbd_core.c
@@ -430,6 +430,21 @@ static inline BOOL USBD_ReqGetDescriptor(void)
                     len = ((USB_STRING_DESCRIPTOR *)pD)->bLength;
                     break;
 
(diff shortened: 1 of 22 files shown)


source/usb/usbd_core.c
```diff
@@ -879,6 +894,9 @@ void USBD_EndPoint0(U32 event)
             case REQUEST_CLASS:
                 switch (USBD_SetupPacket.bmRequestType.Recipient) {
                     case REQUEST_TO_DEVICE:
+                        if (USBD_EndPoint0_Setup_HID_ReqToIF()) {
```
arekzaluski: Instead of creating new hid interface this approach is using device control transfer to send/receive data from the device.

arekzaluski:
Closing temporary as there is a better solution, that used control transfer to interface instead of device.

PR #408, opened 2018-03-14
Improve handling of HID set and get report

c1728p9:
Synchronously handle HID set report so that HID get report is always able to send the result of HID set report. This also simplifies the code and saves rom since the dedicated HID thread is no longer needed.

I tested performance to make sure there were no regressions:

pyOCD programming before this PR:
INFO:root:Programmed 819200 bytes (200 pages) at 45.05 kB/s
INFO:root:Programmed 819200 bytes (200 pages) at 44.99 kB/s
INFO:root:Programmed 819200 bytes (200 pages) at 46.38 kB/s

pyOCD programming with this pr:
INFO:root:Programmed 819200 bytes (200 pages) at 44.70 kB/s
INFO:root:Programmed 819200 bytes (200 pages) at 46.45 kB/s
INFO:root:Programmed 819200 bytes (200 pages) at 46.11 kB/s

### DIFF
--- a/source/daplink/cmsis-dap/usbd_user_hid.c
+++ b/source/daplink/cmsis-dap/usbd_user_hid.c
@@ -38,41 +38,37 @@
 #error "USB HID Input Report Size must match DAP Packet Size"
 #endif
 
-#define FREE_SEM_INIT_COUNT          (DAP_PACKET_COUNT)
-#define PROC_SEM_INIT_COUNT          0
-#define SEND_SEM_INIT_COUNT          0
+#define FREE_COUNT_INIT          (DAP_PACKET_COUNT)
+#define SEND_COUNT_INIT          0
 
-static uint8_t temp_buf                      [DAP_PACKET_SIZE];
 static uint8_t USB_Request [DAP_PACKET_COUNT][DAP_PACKET_SIZE];  // Request  Buffer
 
-static OS_SEM free_sem;
-static OS_SEM proc_sem;
-static OS_SEM send_sem;
+static uint32_t free_count;
+static uint32_t send_count;
 
-static OS_MUT hid_mutex;
-
-// Only used by HID out thread
+// Only used on USB thread
 static uint32_t recv_idx;
-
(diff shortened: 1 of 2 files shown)


c1728p9:
CC @arekzaluski

arekzaluski:
👍

flit:
@c1728p9 Does this fix a specific issue? Also, have you compared performance for the cortex_test.py?

c1728p9:
Hi @flit this allows HID to be used from control endpoints only via HID_REQUEST_SET_REPORT and HID_REQUEST_GET_REPORT. This is useful to @arekzaluski who is looking at adding another USB interface which has no endpoints but provides HID like functionality.

I didn't run cortex_test.py, but I did use flash tool to load a large binary. Performance didn't really change.

flit:
Ok, interesting. I did notice the change where you moved the `USBD_HID_REQ_EP_CTRL` case.

The reason I was asking about cortex_test.py is that I would expect that to be affected more than flash downloads, since it relies more (I think?) on queued packets.

flit:
Aha, I see PR #409… Nice.

c1728p9:
I tried before and after but the values seem dependent more on chance than on this change. Below is the performance on the second consecutive run with new and old firmware.

Cortex test performance with this PR
```
----- TESTING CORTEX-M PERFORMANCE -----
Function getTResponse time: 0.007550
Function step time: 0.010450
Add and remove breakpoint: 0.002850
Function getRegisterContext: 0.006200
Function setRegisterContext: 0.055650
Resume and halt: 0.009450
Simulated GDB step: 0.051250
```

Cortex test performance before this PR
```
----- TESTING CORTEX-M PERFORMANCE -----
Function getTResponse time: 0.008250
Function step time: 0.010800
Add and remove breakpoint: 0.002950
Function getRegisterContext: 0.006950
Function setRegisterContext: 0.056400
Resume and halt: 0.010300
Simulated GDB step: 0.051000
```

PR #409, opened 2018-03-15
Add WebUSB functionality

arekzaluski:
Flash programs on mbed boards using WebUSB: https://developers.google.com/web/updates/2016/03/access-usb-devices-on-the-web

Add new interface - 0xFF (Vendor specific) that it used to send/get hid commands using control transfer to endpoint 0.
In the future this interface will also be used to access serial data.

## Added
- WebUSB descriptor
- WinUSB 2.0 descriptor
- Landing page that points to os.mbed.com

## Tested boards
- BBC micro:bit
- FRDM-K64F
- Nordic nRF51-DK

### Android
- 5.1
- 6.0
* requirements:
  * OTG cable
### Linux
- Ubuntu 16.04
* notes:
  * udev rule modified: https://developers.google.com/web/updates/2016/03/access-usb-devices-on-the-web#tips
### Mac OS
- Mac OS 10.10
- Mac OS 10.12
- Mac OS 10.13
* notes
  * None
### Windows
- Windows 8.1
- Windows 10
* notes
  * landing page is disabled, more info here: https://github.com/WICG/webusb#implementation-status
### Requirements
- Google Chrome version 65+


source/hic_hal/atmel/sam3u2c/usb_config.c
```diff
@@ -28,8 +30,12 @@
 
 //   <o0.0> High-speed
 //     <i> Enable high-speed functionality (if device supports it)
-#define USBD_HS_ENABLE              1
-
+#define USBD_HS_ENABLE              0
```
arekzaluski: I had to disable HS functionality. Windows is showing invalid descriptor message when it is enabled. @c1728p9 any idea what may be wrong?

source/hic_hal/atmel/sam3u2c/usb_config.c
```diff
@@ -347,23 +361,41 @@
 //     </e>
 #define USBD_CLS_ENABLE             0
 
+//     WebUSB support
+#ifdef WEBUSB_INTERFACE
+#define USBD_WEBUSB_ENABLE          1
+#else
+#define USBD_WEBUSB_ENABLE          0
+#endif
+#define USBD_WEBUSB_VENDOR_CODE     0x21
```
arekzaluski: Vendor codes needs to be checked.

flit:
With the WinUSB descriptors, we can support bulk pipes on Windows without requiring a kernel driver.

source/usb/webusb/usbd_core_webusb.c
```diff
@@ -0,0 +1,94 @@
...
+                    if (((WEBUSB_URL_DESCRIPTOR *)pD)->bLength != 0) {
+                        pD += ((WEBUSB_URL_DESCRIPTOR *)pD)->bLength;
+                    }
+                }
+
+                if (((WEBUSB_URL_DESCRIPTOR *)pD)->bLength == 0) {
+                    success = (__FALSE);
+                    break;
+                }
+                strcat(((WEBUSB_URL_DESCRIPTOR *)pD)->URL, "&bid=");
```
arekzaluski: At the moment I've added only board id as parameter in landing page. It's worth considering adding also additional parameters. For example DAPLink version.

devanlai: Since you're including the board id in the query string here, it's probably no longer necessary to include the USB VID - I assume the board id is enough to derive the vendor id for filtering purposes anyways.

For filtering, the best approach would be to pass the serial number in the query string, which the landing page can use to ensure that the USB device picker shows only the device that initiated the landing page notification. Plus the serial number encodes the target id, HIC id, board id, and version anyways.

Since WebUSB is only permitted over HTTPS, I think this should be safe against exposing the user to fingerprinting as long as:

- All outgoing links have the `rel="noreferrer"` attribute set so that they don't leak the serial number via the referrer header
- No untrusted third-party advertising/analytics javascript is included that might log the URL/query string.

arekzaluski: Thanks, I like the idea of extending functionality to send full serial number. I'll make this change.

source/usb/usb_lib.c
```diff
@@ -1654,6 +1780,18 @@ const U8 USBD_DeviceQualifier_HS[] = { 0 };
   HID_REPORT_DESCRIPTOR_TYPE,           /* bDescriptorType */                                               \
   WBVAL(USB_HID_REPORT_DESC_SIZE),      /* wDescriptorLength */
 
+#define WEBUSB_DESC                                                                                            \
+  /* Interface, Alternate Setting 0, VENDOR_SPECIFIC Class */                                                           \
```
c1728p9: nitpick - unaligned slashes

source/usb/usb_lib.c
```diff
@@ -2037,6 +2175,16 @@ const U8 USBD_ConfigDescriptor[] = {
...
 
+#if (USBD_CDC_ACM_ENABLE)
+#if (USBD_MULTI_IF)
+    CDC_ACM_DESC_IAD(USBD_CDC_ACM_CIF_NUM, 2)
+#endif
+    CDC_ACM_DESC_IF0
+    CDC_ACM_EP_IF0
+    CDC_ACM_DESC_IF1
+    CDC_ACM_EP_IF1
+#endif
```
c1728p9: Why was this code moved? If this didn't fix a problem it may be good to move it back to reduce the diff. If it did fix a bug you'll probably want to make the same change for ```USBD_ConfigDescriptor_HS``` later on in this file. @devanlai do you have any insights on this?

source/daplink/cmsis-dap/DAP_vendor.c
```diff
@@ -62,10 +65,47 @@ uint32_t DAP_ProcessVendorCommand(const uint8_t *request, uint8_t *response) {
...
+        num += (read_len + 1);
+        break;
+    }
+    case ID_DAP_Vendor4:  {
+        // uart write
+        int32_t write_len = 64 - 1;
+        uart_write_data((uint8_t *)request, write_len);
+        *response = 1;
+        num += (write_len << 16) | 1;
+        break;
```
c1728p9: @arekzaluski any luck with adding the vendor commands for flashing?

source/hic_hal/atmel/sam3u2c/usb_config.c
c1728p9: This patch makes some changes to the full speed configuration descriptor ```USBD_ConfigDescriptor``` which weren't made to the high speed configuration descriptor ```USBD_ConfigDescriptor_HS```. I left comments below as well in the full speed descriptor asking about this.

source/hic_hal/atmel/sam3u2c/usb_config.c
```diff
@@ -347,23 +361,41 @@
...
+#define USBD_MULTI_IF              (USBD_CDC_ACM_ENABLE*(USBD_HID_ENABLE|USBD_MSC_ENABLE|USBD_ADC_ENABLE|USBD_CLS_ENABLE|USBD_WEBUSB_ENABLE))
 #define USBD_EP_NUM_CALC0           MAX((USBD_HID_ENABLE    *(USBD_HID_EP_INTIN     )), (USBD_HID_ENABLE    *(USBD_HID_EP_INTOUT!=0)*(USBD_HID_EP_INTOUT)))
 #define USBD_EP_NUM_CALC1           MAX((USBD_MSC_ENABLE    *(USBD_MSC_EP_BULKIN    )), (USBD_MSC_ENABLE    *(USBD_MSC_EP_BULKOUT)))
 #define USBD_EP_NUM_CALC2           MAX((USBD_ADC_ENABLE    *(USBD_ADC_EP_ISOOUT    )), (USBD_CDC_ACM_ENABLE*(USBD_CDC_ACM_EP_INTIN)))
 #define USBD_EP_NUM_CALC3           MAX((USBD_CDC_ACM_ENABLE*(USBD_CDC_ACM_EP_BULKIN)), (USBD_CDC_ACM_ENABLE*(USBD_CDC_ACM_EP_BULKOUT)))
 #define USBD_EP_NUM_CALC4           MAX(USBD_EP_NUM_CALC0, USBD_EP_NUM_CALC1)
 #define USBD_EP_NUM_CALC5           MAX(USBD_EP_NUM_CALC2, USBD_EP_NUM_CALC3)
 #define USBD_EP_NUM_CALC6           MAX(USBD_EP_NUM_CALC4, USBD_EP_NUM_CALC5)
-#define USBD_EP_NUM                (USBD_EP_NUM_CALC6)
+#define USBD_EP_NUM                 USBD_EP_NUM_CALC6
```
c1728p9: nitpick - if you leave the parentheses it will reduce the diff

source/usb/usb_lib.c
```diff
@@ -2281,11 +2438,87 @@ const struct {
...
+#if (USBD_WEBUSB_ENABLE)
+
+#define WEBUSB_NUM_FUNCTIONS              (1)
+
+#define WEBUSB_WTOTALLENGTH               (WEBUSB_DESCRIPTOR_SET_HEADER_SIZE +      \
+                                           WEBUSB_CONFIGURATION_SUBSET_HEADER_SIZE +\
+                                           (WEBUSB_NUM_FUNCTIONS * (WEBUSB_FUNCTION_SUBSET_HEADER_SIZE + 1)))
+
+__weak \
+const U8 USBD_WebUSBAllowedOriginsHeader[] = {
```
devanlai: You may want to remove the allowed origins descriptor, since it was dropped from the WebUSB spec. It's benign to leave it in, but it's less code to maintain.

source/usb/usb_lib.c
```diff
@@ -1633,6 +1658,107 @@ __weak \
...
+	USB_DEVICE_CAPABILITY_DESCRIPTOR_TYPE,  /* bDescriptorType */
+	USB_DEVICE_CAPABILITY_PLATFORM,         /* bDevCapabilityType */
+	0x00,                                   /* bReserved */
+	0x38, 0xB6, 0x08, 0x34,                 /* PlatformCapabilityUUID */
+	0xA9, 0x09, 0xA0, 0x47,
+	0x8B, 0xFD, 0xA0, 0x76,
+	0x88, 0x15, 0xB6, 0x65,
+	WBVAL(0x0100), /* 1.00 */               /* bcdVersion */
+	USBD_WEBUSB_VENDOR_CODE,                /* bVendorCode */
+	1,                                      /* iLandingPage */
```
devanlai: I'm not sure if it would be worth the effort, but if you wanted to be able to disable the landing page popup (which should really be handled by the user-agent), you could override `iLandingPage` to 0 at runtime.

source/usb/usb_lib.c
devanlai: I'm out of time to review it further this morning, but Windows is stricter about devices that report a USB 2.0 (as is required for the WebUSB Binary Object Store descriptor). If the interface numbers don't match the order in which the interfaces appear in the configuration descriptor, Windows will report an error. This may be related to the HS interface not functioning.

source/usb/usb_lib.c
```diff
@@ -2281,11 +2438,87 @@ const struct {
...
 #endif
+#if (USBD_WEBUSB_ENABLE)
+    USBD_STR_VAL(WEBUSB_STRDESC),
+#endif
 #if (USBD_MSC_ENABLE)
     USBD_STR_VAL(MSC_STRDESC),
 #endif
 };
 
+#if (USBD_WEBUSB_ENABLE)
```
microbit-carlos: Trying to test this I was building the micro:bit DAPLink bootloader (kl26z_bl) on this branch, and in my environment this `if` macro ends up evaluating to `true`, so the `else` in line 2510 isn't compiled and the `USBD_EndPoint0_Setup_WebUSB_ReqToDevice ` symbol is never resolved:
```
*** Using Compiler 'V5.06 update 5 (build 528)', folder: 'C:\Keil_v5\ARM\ARMCC\Bin'
Build target 'kl26z_bl'
Before Build - User command #1: ..\..\..\tools\pre_build_script.bat
(env) C:\workspace\microbit-foundation\daplink-websub\projectfiles\uvision\kl26z_bl>python ..\..\..\tools\pre_build_script.py 
#> Pre-build script start
#> Getting git SHA
#> Checking for local changes
#> Keeping git version file since it didn't need to change
#> Pre-build script completed
compiling usb_config.c...
linking...
.\build\kl26z_bl.axf: Error: L6218E: Undefined symbol USBD_EndPoint0_Setup_WebUSB_ReqToDevice (referred from usbd_core.o).
Not enough information to list image symbols.
Not enough information to list load addresses in the image map.
Finished: 2 information, 0 warning and 1 error messages.
".\build\kl26z_bl.axf" - 1 Error(s), 0 Warning(s).
Target not created.
```
Can you replicate when building the kl26z_bl? Is it an issue on my configuration, or is it a known issue?

microbit-carlos:
This is shaping up great, It's really exciting to test the demo from https://armmbed.github.io/dapjs/examples/web.html!

I was testing the `arekzaluski:f/webusb_hid_serial` branch and noticed that on first connection to Chrome the displayed named for the micro:bit was "LPC1768", but once it is connected and I restart the browser it changes to "DAPLink CMSIS-DAP":
image

source/usb/usb_lib.c
arekzaluski: I'm disabling the landing page functionality for now. At least until it is officially supported again in Chrome on Windows.

source/usb/webusb/usbd_core_webusb.c
arekzaluski: We discussed it and decided to remove landing page functionality for now. I'm going to remove it.

source/usb/usb_lib.c
arekzaluski: You are correct. It's no longer in specification `0x01` is now marked as reserved and `0x02` is used to get url. I'm removing it.

arekzaluski:
I've updated this PR based on everyone's comments. Big thanks for the feedback. Changes:
- Fixed support for high speed feature
- Disabled landing page ( Can be easily brought back later when it will be fully implemented in Chrome in Windows)
- Removed deprecated allowed origins descriptor 
- Removed vendor id from landing page
- Cleaned the code

Flashing the program on the board and reading/writing serial data using WebUSB can be tested on https://armmbed.github.io/dapjs

source/usb/usb_lib.c
arekzaluski: @microbit-carlos Thanks for pointing it out. I've just fixed it for bootloaders.

thegecko:
Great stuff, @arekzaluski @c1728p9 Can we merge this into master? cc @jaustin

JoeTheGuitarist:
Great stuff, thanks for your hard work on this @arekzaluski and @c1728p9 for your guidance. Looking forwards to seeing this merged!

PR #412, opened 2018-03-27
Bump ram size on RAK811

c1728p9:
Bump ram size to match hardware.


ithinuel:
The actual part is **STM32L151CBU6A**.
This chip revision embeds 32KiB of ram instead of only 16KiB for the STM32L151CBU6 (without the **A**).

PR #419, opened 2018-04-11
Update DAPLink documentation

linlingao:

PR #420, opened 2018-04-11
Fix Windows testing failures due to indexing

c1728p9:
Add a delay before copying files on to DAPLink to allow Windows 8 and Windows 10 to complete any indexing it needs to do when a device is connected.


c1728p9:
With this workaround the need for turning off drive indexing is no longer needed in #286 and #419.

CC: @deepikabhavnani @linlingao

PR #422, opened 2018-04-18
Rapid-IoT: Update KW41 Flash Algo

mmahadevan108:
Failures were seen with the old flash algo when erasing sectors
Update MDM_ID based on K64 or KW4

Signed-off-by: Mahesh Mahadevan <mahesh.mahadevan@nxp.com>


mmahadevan108:
cc @c1728p9 @jasonNXP

source/target/freescale/rapid_iot/target_reset.c
```diff
@@ -85,7 +92,7 @@ uint8_t target_unlock_sequence(void)
     }
 
     // verify the result
-    if (val != MDM_ID) {
+    if (val != mdm_id) {
```
c1728p9: Since this code is the same for both cases you could check for either. What are you thoughts on something like this:
```C
if ((val != MDM_ID_K64)  && (val != MDM_ID_KW4 )) {
```
What you have is good as well, so don't feel like you need to change it.

c1728p9:
This is an internal failure unrelated to your change @mmahadevan108. I'm working on getting this resolved now.

mmahadevan108:
Thanks Russ, yes it fixes an issue where code was looking for the wrong ID when connected to the KW4.

mmahadevan108:
@c1728p9 Can this be merged to master.

PR #425, opened 2018-04-26
Rapid-IoT: Update implementation to HALT target

mmahadevan108:
When the Kinetis flash is empty the reset line toggles. This causes failures when trying to halt the target. Use the reset halt method in this case.

### DIFF
--- a/source/target/freescale/rapid_iot/target_reset.c
+++ b/source/target/freescale/rapid_iot/target_reset.c
@@ -54,7 +54,13 @@ void prerun_target_config(void)
     mdm_id = MDM_ID_K64;
 
     // get a hold of the target
-    target_set_state(HALT);
+    if (target_set_state(HALT) == 0) {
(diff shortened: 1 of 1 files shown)


mmahadevan108:
cc @c1728p9

source/target/freescale/rapid_iot/target_reset.c
```diff
@@ -75,7 +83,11 @@ void prerun_target_config(void)
     }
 
     // Let the target run
-    target_set_state(RUN);
+    if (reset_program) {
+        target_set_state(RESET_RUN);
+    } else {
+        target_set_state(RUN);
```
c1728p9: Is there a problem with using RUN in both cases?

mmahadevan108:
Can this PR be merged to master. I would like to give the team an updated binary from master branch with this fix included for their testing.

Issue #430, opened 2018-05-11
Feature Request: Target URL in details.txt


microbit-carlos:
Hi all,

For the micro:bit we've got a couple of use cases where it would be quite beneficial for some tools and editors to be able to easily parse the project URL from the micro:bit itself.
At the moment the URL is included in the generated HTML, but parsing through that is a bit painful and far from ideal, so we were hoping the details.txt file could be updated to include this URL as well.

Could we get a feature like this included in DAPLink?

microbit-carlos:
I've prototyped this in PR https://github.com/ARMmbed/DAPLink/pull/431.

microbit-carlos:
This was implemented in PR https://github.com/ARMmbed/DAPLink/pull/460.

PR #431, opened 2018-05-11
Feature request PR: Add target URL to details.txt file

microbit-carlos:
As discussed in issue #430.

I'm not very content to use the `file_buffer` statically allocated array to temporary store the URL before it is copied over, but at this level we don't know how long the URL will be until it is "extended", and the contents of this buffer are not really used for anything other than to calculate file sizes in the `get_file_size()` function.

Suggestions and improvements are very welcomed, even if they mean doing this completely differently.

### DIFF
--- a/source/daplink/drag-n-drop/vfs_user.c
+++ b/source/daplink/drag-n-drop/vfs_user.c
@@ -75,6 +75,8 @@ static uint32_t read_file_assert_txt(uint32_t sector_offset, uint8_t *data, uint
 static uint32_t read_file_need_bl_txt(uint32_t sector_offset, uint8_t *data, uint32_t num_sectors);
 
 static void insert(uint8_t *buf, uint8_t *new_str, uint32_t strip_count);
+static uint32_t insert_info(uint8_t *buf, uint32_t bufsize, const char *content, uint32_t content_size);
+static uint32_t insert_target_url(uint8_t *buf, uint32_t bufsize);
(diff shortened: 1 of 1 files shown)


jaustin:
@c1728p9 any comments on Carlos's RFC here?

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -310,6 +312,13 @@ static uint32_t read_file_details_txt(uint32_t sector_offset, uint8_t *data, uin
     pos += util_write_string(buf + pos, "Remount count: ");
     pos += util_write_uint32(buf + pos, remount_count);
     pos += util_write_string(buf + pos, "\r\n");
+
+    // Target URL, use file_buffer as a temporary storage
+    insert_target_url(file_buffer, ARRAY_SIZE(file_buffer));
```
brianesquilona: Because of line 95, the file_buffer will be re/used twice which may lead to incorrect computation on some cases.

brianesquilona:
This PR will not be merged, but we will going to rewrite the code generation for details.txt to include the URL

microbit-carlos:
Great, will the feature make it to the next DAPLink release?

brianesquilona:
@microbit-carlos  Yes, it will be available on the next release.

microbit-carlos:
Implemented in https://github.com/ARMmbed/DAPLink/pull/460.

PR #432, opened 2018-05-16
Add sector erase test

linlingao:


test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
     #        when a binary file is loaded, since there is no way to
     #        tell where the end of the file is.
 
+    if page_erase_supported:
```
flit: I think it would be good to extract the page erase tests to another function, to keep the `test_mass_storage()` function from growing too large.

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
+            memory_map = mbed_board.target.getMemoryMap()
+        flash_regions = [region for region in memory_map if region.type == 'flash']
+        number_flash_regions = len(flash_regions)
+
+        max_address = intel_hex.maxaddr()
+        # Create an object. We'll add the addresses of unused even blocks to it first, then unused odd blocks for each region
+        ih = intelhex.IntelHex()
+        # Add the content from test bin first
+        expected_bin_contents = bin_file_contents
+        for region_index in range(0, number_flash_regions, 1):
```
flit: Fyi, the *step* parameter is not required if it is set to 1, as that is the default.

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
+            flash_start = flash_regions[region_index].start
+            flash_length = flash_regions[region_index].length
+            block_size = flash_regions[region_index].blocksize
+            number_of_blocks = int(flash_length/block_size)
```
flit: Use the `//` operator instead of the `int` constructor.

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
+
```
flit: Might want to add an assertion to verify that all flash regions are contiguous. We don't need to handle testing with discontiguous flash regions right now (presumably we don't even have any such devices), but it would be good to force a failure if such a device appears since the test won't work right.

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
+            if max_address >= (flash_start + flash_length):
+                # This bin image crosses this region, don't modify the content, go to the next region
+                continue
+            elif max_address >= flash_start:
+                # This bin image occupies partial region. Skip the used portion to avoid touching any security bits and pad the rest
+                expected_bin_contents += bytearray([0xff]) * (flash_start + flash_length - max_address - 1)
+                # Calculate the starting block after the image to avoid stumbling upon security bits
+                block_start = int((max_address - flash_start)/block_size) + 1
```
flit: `//` operator.

test/msd_test.py
linlingao: This PR is getting big. Let me improve this in another PR because I need to another negative test to override the blocksize to be smaller to deliberately make the test fail.

test/msd_test.py
linlingao: Got it, will remove the step parameter.

test/msd_test.py
linlingao: Got it, will change. Thanks for letting me know.

test/msd_test.py
linlingao: Good idea. Will add in the next commit.

test/msd_test.py
linlingao: Yep

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
 
+    if page_erase_supported:
+        # Test page erase, a.k.a. sector erase by generating iHex with discrete addresses,
+        # programing the device then comparing device memory against expected content.
+        test = MassStorageTester(board, test_info, "Sector Erase")
+        with MbedBoard.chooseBoard(board_id=board.get_unique_id(), init_board=False) as mbed_board:
+            memory_map = mbed_board.target.getMemoryMap()
+            flash_regions = [region for region in memory_map if region.type == 'flash']
+            number_flash_regions = len(flash_regions)
+            mbed_board.uninit(resume=False, disconnect=False)
```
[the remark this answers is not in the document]
linlingao: Done.

test/msd_test.py
flit: That's fine.

test/msd_test.py
```diff
@@ -373,6 +375,56 @@ def test_mass_storage(workspace, parent_test):
...
+            max_address = intel_hex.maxaddr()
+            # Create an object. We'll add the addresses of unused even blocks to it first, then unused odd blocks for each region
+            ih = intelhex.IntelHex()
+            # Add the content from test bin first
+            expected_bin_contents = bin_file_contents
+            for region_index in range(0, number_flash_regions, 1):
```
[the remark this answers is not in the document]
linlingao: Thanks, will change.

test/msd_test.py
```diff
@@ -373,6 +375,63 @@ def test_mass_storage(workspace, parent_test):
...
+                    modifier = 0x55
+                else:
+                    modifier = 0xAA
+                    block_start += 1
+                for block_idx in range(block_start, number_of_blocks, 2):
+                    for address_to_modify in range (flash_start + block_idx * block_size, flash_start + (block_idx + 1) * block_size):
+                        expected_bin_contents[address_to_modify] = modifier
+                        ih[address_to_modify] = modifier
+        if not os.path.exists("tmp"):
+            os.makedirs("tmp")
```
flit: (Not in this PR, but soon…) Take a look at the Use the built-in `tempfile` Python package to create temp files in the system temp directory instead of creating your own local "tmp" dir.

linlingao: Got it, thanks for letting me know.

flit:
@linlingao It's failing due to the Linux serial port problem…

PR #436, opened 2018-05-30
Enable sector erase on Nordic nRF5x targets

linlingao:


source/board/mkit_dk_dongle_nrf5x.c
```diff
@@ -22,6 +22,8 @@
 #include "sam3u2c.h"
 #include "target_config.h"
 #include "util.h"
+#include "stdbool.h"
```
flit: Instead of adding the `stdbool.h` include here, add it to `flash_manager.h` where it's missing. Headers should (usually) be standalone and include all required includes within them.

linlingao: Sure, will incorporate it in the next commit.

source/target/nordic/nrf5x/flash_blob.c
```diff
@@ -44,6 +44,14 @@ static const uint32_t nRF52832AA_FLM[] = {
     0x40010404, 0x40010504, 0x6e524635, 0x00000000, 
 };
 
+static const sector_info_t sectors_info[] = {
```
flit: Fyi, you don't need to add `sector_info_t` structures unless you have multiple sector sizes. The code in `target_flash_erase_sector_size()` will return `target_device.sector_size` if `target_device.sector_info_length` is 0. And that's the only use of this sector size info.

linlingao: target_device.sector_size and target_device.sector_cnt are removed if sector_info[] is added.

test/msd_test.py
```diff
@@ -398,7 +399,7 @@ def test_mass_storage(workspace, parent_test):
 
             # Sanity check the regions are contiguous
             if region_index:
```
flit: Minor note: Use a "> 0" comparison since, semantically, `region_index` is not a Boolean value.

test/msd_test.py
```diff
@@ -382,14 +382,15 @@ def test_mass_storage(workspace, parent_test):
...
         flash_regions = [region for region in memory_map if region.type == 'flash']
-        number_flash_regions = len(flash_regions)
 
         max_address = intel_hex.maxaddr()
         # Create an object. We'll add the addresses of unused even blocks to it first, then unused odd blocks for each region
         ih = intelhex.IntelHex()
         # Add the content from test bin first
         expected_bin_contents = bin_file_contents
         for region_index, the_region in enumerate(flash_regions):
+            if the_region.isBootMemory == False:
```
flit: Very minor note: use `not the_region.isBootMemory` since it reads better.

(Fyi, you don't need to change the minor notes right now, they are just suggestions for more readable code.)

source/target/nordic/nrf5x/flash_blob.c
flit: Right, but you can simplify the code by keeping `sector_size` and `sector_cnt` and not having `sector_info`. Fewer changes, too.

(Btw, `sector_cnt` doesn't appear to be used anywhere in the code!)

source/board/mkit_dk_dongle_nrf5x.c
linlingao: Committed.

source/target/nordic/nrf5x/flash_blob.c
linlingao: Since some targets have multiple sector sizes, would it be better to use the same data structure, sector_info?

test/msd_test.py
linlingao: Minor thought: I usually prefer positive logic. In my mind, something == false is more readable than !something. It's okay either way.

flit: Understandable :)

source/target/nordic/nrf5x/flash_blob.c
flit: Generally yes, but it is already the case that most targets use `sector_size`. So unless we make a global change to `sector_info`, there will be a mix. Maybe we can do this when we switch to Pack-based algos. For now, I think it's best to limit changes unless required.

source/target/nordic/nrf5x/target.c
```diff
@@ -26,34 +26,34 @@
```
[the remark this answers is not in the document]
linlingao: Done

PR #438, opened 2018-05-31
Fix interfaces order.

arekzaluski:
USB descriptor is broken on Windows 10 machine after latest WebUSB Pull request. It is because order of interfaces is important on Windows machine. This PR fixes this behaviour.

@c1728p9


flit:
What are the requirements on interface ordering for Windows 10? Or where is it documented? (Mostly for my own edification.)

jaustin:
@arekzaluski any idea what the CI fail was here?

c1728p9:
@jaustin I think this was a spurious failure. I'll re-trigger.

PR #450, opened 2018-07-03
support for lpc11u35 dipdap sdt boards

brianesquilona:
boards added: sdt32429b sdt32439b sdt64b sdt51822b sdt52832b


jwyune:
Brian (@brianesquilona), I think std64b is supposed to be sdt64b.

brianesquilona:
@jwyune you are correct. I have updated the affected files.

flit:
@brianesquilona Looks really good! The only thing missing is adding the board IDs to the `BOARD_ID_SUPPORTING_PAGE_ERASE` list in `info.py`.

test/info.py
```diff
@@ -231,6 +231,11 @@
     0x0311,  # K66F
     0x1101,  # Nordic-nRF52-DK
     0x1102,  # Nordic-nRF52840-DK
+    0x3104,  # dipdap_sdt52832b
+    0x3108,  # dipdap_sdt32429b
+    0x3110,  # dipdap_sdt32439b
+    0x3105,  # dipdap_sdt64b
+    0x3103,  # dipdap_sdt51822b
```
flit: Sorry… can you please keep the list sorted by ID? (This applies to any list in DAPLink and pyOCD that has IDs as the primary key.) Thanks!

PR #451, opened 2018-07-06
Add support for GR-LYCHEE

d-kato:


source/board/gr-lychee.c
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "stdbool.h"
+#include "virtual_fs.h"
+#include "flash_manager.h"
+
+const char *board_id = "5501";
+
+const vfs_filename_t daplink_drive_name =     "MBED       ";
```
flit: This drive name should not be required since "MBED" is the default. If you do leave it, you need to add `__attribute__((aligned(4)))` as a workaround for an armcc bug with weak references.

source/target/renesas/rza1lu/gr-lychee/validation_gr-lychee.c
```diff
@@ -0,0 +1,48 @@
+/**
```
flit: Looks like this file is the same as `validation_gr-peach.c`. Please rename the one for gr-peach and reuse it instead of adding a duplicate source file.

test/info.py
```diff
@@ -173,6 +174,7 @@
     (   0x7404,     'lpc11u35_6lowpan_borderrouterethernet_if', None,               'mbed-6LoWPAN-Border-Router-ETHERNET'   ),
     (   0xC000,     'lpc11u35_cocorico_if',                     None,               'CoCo-ri-Co'                            ),
     (   0x5500,     'lpc11u35_gr_peach_if',                     None,               None                                    ), # TODO - Set to 'Renesas-GR-PEACH' once this target builds
+    (   0x5501,     'lpc11u35_gr_lychee_if',                    None,               None                                    ), # TODO - Set to 'Renesas-GR-LYCHEE' once this target builds
```
flit: Please add this board ID to the `BOARD_ID_SUPPORTING_PAGE_ERASE` list near the bottom of `info.py`, since you have enabled page erase on this board.

source/board/gr-lychee.c
d-kato: The default driver name is "DAPLINK". So I added "__attribute__((aligned(4)))".

flit:
Regarding the ci-morph-test failure: only the Linux test is failing due to an apparent configuration issue on the test host.

studavekar:
> Regarding the ci-morph-test failure: only the Linux test is failing due to an apparent configuration issue on the test host.

This is fixed.

PR #456, opened 2018-07-14
IO_Config.h: Modify the way it is included

findlayfeng:
This header file defines gpio and pin,
but is included in many places that are not needed.
Because these places need to introduce cmsis_core,
and because cmsis_core depends on the IRQn_Type definition of soc,
the cmsis_core header file cannot be directly imported in the corresponding place.
So modify the response of the .yaml file with the --preinclude parameter
to globally include the definition of the soc containing the cmsis_core header file.
"#include "IO_Config.h"" is safely removed.
Added only where gpio and pin definitions are needed

Signed-off-by: Findlay Feng <i@fengch.me>


cederom:
Not sure here, someone with better insight should analyze/comment plz? :-)

brianesquilona:
The intention is good, but I am not sure about adding an specific path of a file in a yaml, will make shifting to other build environment less flexible. I will close this PR soon if there are no objections.

0xc0170:
> The intention is good, but I am not sure about adding an specific path of a file in a yaml, will make shifting to other build environment less flexible. I will close this PR soon if there are no objections.

This will require further work. I'll close this for now.

PR #457, opened 2018-07-14
tools: build_script: python version

findlayfeng:
I also installed python2 and python3, my python3 is more commonly used, python is my default point to python3, causing my script to fail
Modify the test py command to exist, if it exists, use py -2 to call python2 version

Signed-off-by: Findlay Feng <i@fengch.me>

### DIFF
--- a/tools/build_release_uvision.bat
+++ b/tools/build_release_uvision.bat
@@ -30,6 +30,13 @@ cd %~dp0..\
 )
 @echo USING UV4=%UV4%
 
+@where py 1> nul
+@if "%errorlevel%"=="0" (
+    set PYTHON_CMD=py -2
+) else (
+    set PYTHON_CMD=python
+)
+
 rmdir /q /s uvision_release
 @if %errorlevel% neq 0 exit /B %errorlevel%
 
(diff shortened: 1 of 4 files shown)


findlayfeng:
First %errorlevel% is not defined
He is the result of the execution of the previous statement
Use where to test whether py exists, if it exists, it is convenient to specify the py version.
Https://docs.python.org/3.7/using/windows.html#python-launcher-for-windows
Python --version does not return version 2, does not mean that python2 is not installed, requires more testing
Well, here I think I need to add 2> nul will hide the hint, next Monday

findlayfeng:
If python2 is installed, python3 is not installed, py does not exist, use python directly
If python2 is installed, python3 is also installed, py exists, use py -2
If python2 is not installed, python3 is installed. If py -2 fails, the script itself will fail, but such a failure will obviously show that python2 is not installed.

findlayfeng:
Well, I use Google Translate, it may not be very clear

cederom:
I think only Python2 is currently supported anyway.. why don't we simply use `python2`?

findlayfeng:
Python3 is future, or should be used first.
Also no python2.exe in windows
I think the best way is to specify the version at the beginning of the .py file and open the .py file directly instead of using the python command.

flit:
Imo, the proper solution here is just to support Python 3 in the build and test scripts.

brianesquilona:
#499 should be able support more than one virtualenv both for python 2 and 3, tools and tests. Closing this PR.

PR #460, opened 2018-07-24
Added support for URL in DETAILS.TXT

brianesquilona:
Added a overflow check on 512 buffer after expanding special characters


flit:
Note: Linux test failure is caused by a known intermittent failure related to serial only on Linux.

PR #461, opened 2018-07-25
microbit: Autodetect board type on runtime

microbit-carlos:
Pin 20 on the micro:bit KL26 has a no-fit part for an external-pull in v1.3 of the board (so it is left floating), and this external pull up is fitted in v1.5.
This PR sets the internal pull down on pin20, identifies the board and sets the respective `board_id ` based on the pin20 digital read.

### DIFF
--- a/source/board/microbit.c
+++ b/source/board/microbit.c
@@ -18,8 +18,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-const char *board_id = "9900";
+#include "MKL26Z4.h"
+#include "IO_Config.h"
 
 // URL_NAME and DRIVE_NAME must be 11 characters excluding
 // the null terminated character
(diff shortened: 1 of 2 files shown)


source/board/microbit.c
```diff
@@ -18,8 +18,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-const char *board_id = "9900";
+#include "MKL26Z4.h"
```
flit: You don't have to change anything, but for future reference it is preferable to include `fsl_device_registers.h` instead of the specific register header by name.

source/board/microbit.c
```diff
@@ -30,3 +30,47 @@ __attribute__((aligned(4)))
...
+} mb_version_t;
+
+// Enables Board Type Pin, reads it and disables it
+// Depends on gpio_init() to have been executed already
+uint8_t read_board_type_pin(void) {
+    uint8_t pin_state = 0;
+    // GPIO mode, Pull enable, pull down, input
+    PIN_BOARD_TYPE_PORT->PCR[PIN_BOARD_TYPE_BIT] = PORT_PCR_MUX(1) | PORT_PCR_PE(1) | PORT_PCR_PS(0);
+    PIN_BOARD_TYPE_GPIO->PDDR &= ~PIN_BOARD_TYPE;
+    // Read pin
```
flit: You might need to add a small delay here after you configure the internal pull-down, before you read the pin. The internal pull-downs are fairly weak at 20-50 kΩ.

microbit-carlos:
@flit I'm not sure how the tests are run in the CI, are they running in an emulator or are boards attached to a server?

flit:
@microbit-carlos The boards are attached directly to servers for each of the three OSes.

source/board/microbit.c
```diff
@@ -30,3 +30,49 @@ __attribute__((aligned(4)))
...
+
+// Enables Board Type Pin, reads it and disables it
+// Depends on gpio_init() to have been executed already
+uint8_t read_board_type_pin(void) {
+    uint8_t pin_state = 0;
+    // GPIO mode, Pull enable, pull down, input
+    PIN_BOARD_TYPE_PORT->PCR[PIN_BOARD_TYPE_BIT] = PORT_PCR_MUX(1) | PORT_PCR_PE(1) | PORT_PCR_PS(0);
+    PIN_BOARD_TYPE_GPIO->PDDR &= ~PIN_BOARD_TYPE;
+    // Wait to stabilise, based on gpio.c busy_wait(), at -O2 10000 iterations delay ~850us
+    for (volatile uint32_t i = 10000; i > 0; i--);
```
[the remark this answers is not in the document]
flit: Should be long enough a delay. The RC time constant for 50 kΩ and the nominal pin capacitance of 7 pF is only 0.3 ns.

flit:
@microbit-carlos I'm afraid the last test run still incorrectly reported the board ID as 9901. I'm going to take a look at the boards to see if they have that resistor populated.

PR #472, opened 2018-08-06
Use this patch for current yml format for gh-pages

brianesquilona:


tools/make_update_yml.py
```diff
@@ -0,0 +1,179 @@
...
+            '   - `cp <path to firmware file> <CRP DISABLD> && sync`',
+            '   - Note: make sure to change `CRP DISABLD` to the name of the mount point on your system.',
+            '5. Power cycle the board. It will now enumerate and mount as `DAPLINK` or the name of the board.'
+        ])
+    ])
+
+}
+
+
+def string_writer(element):
```
flit: Indent spacing is off from here to the end of the file. Should be 4 spaces per indent.

tools/make_update_yml.py
```diff
@@ -0,0 +1,179 @@
...
+	elif type(element) is int:
+		return "_0x%04i" % element
+	else:
+		return None
+
+
+def log_string (exeption, type, element):
+	print "line: " + str(currentframe().f_back.f_lineno) + " " +  exeption +  " " + str(type) +"\n in element: "+ element 
+
+def yml_object_parser(f,entry,level,tabs):
```
flit: Formatting: gotta watch those spaces after commas and around operators. 😉 Please check spacing on all changes (there are a few more in this file, as well as `package_releaes_files.py`—I won't enumerate all of them.

tools/make_update_yml.py
```diff
@@ -0,0 +1,179 @@
...
+def string_writer(element):
+	if type(element) is str:
+		return element
+	elif type(element) is int:
+		return "_0x%04i" % element
+	else:
+		return None
+
+
+def log_string (exeption, type, element):
```
flit: Instead of defining your own log function, use the `logging` package.

Comments that don't matter given the previous sentence: "exeption" is misspelled ("exception"). I would also use a different term than "exception" since that implies a real exception object, where the parameter here is an error message.

tools/make_update_yml.py
brianesquilona: Thank Chris, I have just updated the files

Issue #467, opened 2018-08-02
DAPLink flashing isn't working after hardfault crash


KariHaapalehto:
Step to procedure error:
- Create mbed-os version, which will hardfault e.g Null pointer write. Flash the sw and verify from traces that hardfault did happen.
- Correct you sw and try to flash it. HW will blink the led just couple times and the "FAIL.TXT"-file is created to daplink-drive with following text:
"error: The interface firmware FAILED to reset/halt the target MCU
type: target"
and no new is flashed.

-Only way to recover from this issue is manually press reset button down before sw copy starts and release the reset button while coping.

This error has been seen with several new MTB+MCB compination. I have seeen it with MTB_USI_BN_BM_22, MTB_ADV_WISE_1530 and MTB_MXCHIP_EMW3166. 
I have seen this error with DAPLink versions 0246 and 0244.

brianesquilona:
Fixed in the PR #480

PR #480, opened 2018-08-15
Try to do a rewrite when AP fails after a reset.

brianesquilona:
MTB_ADV_WISE_1530 issues when the target is in invalid state, AP writes fail. 
#467


flit:
Do you know why the target gets into an invalid state, and more specifically, what is causing the writes to fail?

PR #488, opened 2018-08-28
Added option to customize usb product string

brianesquilona:
Tested with uVision and IAR to have no effect on debugger (ids and download).


jaustin:
Was this a legit failure, or CI issue?

studavekar:
test passed need to use the latest version of mbed-ls which is released a couple of hours ago : https://github.com/ARMmbed/mbed-ls/releases/tag/release-version-1.6.0

PR #497, opened 2018-09-10
Size can zero in browser download, any size transferred can be a file optional finish

brianesquilona:
EDGE download has this symptom.

### DIFF
--- a/source/daplink/drag-n-drop/vfs_manager.c
+++ b/source/daplink/drag-n-drop/vfs_manager.c
@@ -76,7 +76,8 @@ typedef enum {
 
 typedef struct {
     vfs_file_t file_to_program;     // A pointer to the directory entry of the file being programmed
-    vfs_sector_t start_sector;      // Start sector of the file being programmed
+    vfs_sector_t start_sector;      // Start sector of the file being programmed by stream
+    vfs_sector_t file_start_sector; // Start sector of the file being programmed by vfs
     vfs_sector_t file_next_sector;  // Expected next sector of the file
     vfs_sector_t last_ooo_sector;   // Last out of order sector within the file
     uint32_t size_processed;        // The number of bytes processed by the stream
@@ -100,6 +101,7 @@ typedef enum {
 
 static const file_transfer_state_t default_transfer_state = {
     VFS_FILE_INVALID,
+    VFS_INVALID_SECTOR,
     VFS_INVALID_SECTOR,
     VFS_INVALID_SECTOR,
     VFS_INVALID_SECTOR,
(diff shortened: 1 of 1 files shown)


source/daplink/drag-n-drop/vfs_manager.c
```diff
@@ -615,12 +615,15 @@ static void transfer_update_file_info(vfs_file_t file, uint32_t start_sector, ui
...
 {
     vfs_mngr_printf("vfs_manager transfer_reset_file_info()\r\n");
-    if (file_transfer_state.stream_open) {
-        transfer_update_state(ERROR_ERROR_DURING_TRANSFER);
-    } else {
-        file_transfer_state = default_transfer_state;
-        abort_remount();
-    }
+    //check if the data started streaming; size can be updated on matching start sector and stream type
+    if(file_transfer_state.stream_started && (TRASNFER_FINISHED != file_transfer_state.transfer_state)){
```
c1728p9: You should only need to check `stream_started` here, as it only gets reset back to 0 on remount.

source/daplink/drag-n-drop/vfs_manager.c
```diff
@@ -737,6 +740,7 @@ static void transfer_update_state(error_t status)
     // 1. A file has been detected
     // 2. The size of the file indicated in the root dir has been transferred
     // 3. The file size is greater than zero
+    // 4. Size can zero like in EDGE browser download, in this case any  size transferred can be a file optional finish
```
c1728p9: Is this comment still relevant?

source/daplink/drag-n-drop/vfs_manager.c
```diff
@@ -615,12 +615,15 @@ static void transfer_update_file_info(vfs_file_t file, uint32_t start_sector, ui
...
-    if (file_transfer_state.stream_open) {
-        transfer_update_state(ERROR_ERROR_DURING_TRANSFER);
-    } else {
-        file_transfer_state = default_transfer_state;
-        abort_remount();
-    }
+    //check if the data started streaming; size can be updated on matching start sector and stream type
+    if(file_transfer_state.stream_started && (TRASNFER_FINISHED != file_transfer_state.transfer_state)){
+        //file to program can change but all fields should be preserved
+        file_transfer_state.file_to_program = VFS_FILE_INVALID;
```
c1728p9: You'll probably want to reset `file_size` and `start_sector` as well, since these may be set to the wrong value from the deleted file.

The value `start_sector` is a bit problematic, since the stream or root directory could have written this. One way to safely handle this is to create two variables, one for the start sector detected by the stream and one for the root directory - maybe something like `start_sector` and `file_start_sector`. Then you could reset `file_start_sector` here while leaving `start_sector` unchanged.  You could then make use of this as part of the optional finish check:
```C
    file_transfer_state.file_info_optional_finish =
        (file_transfer_state.file_to_program != VFS_FILE_INVALID) &&
        (file_transfer_state.size_transferred >= file_transfer_state.file_size) &&
        (file_transfer_state.file_size > 0) &&
        (file_transfer_state.start_sector == file_transfer_state.file_start_sector);
```

brianesquilona:
@c1728p9 Thanks for the review. I have updated the code to support the sector size split.

brianesquilona:
daplink_build_pr 326
daplink_test_pr 307

PR #503, opened 2018-09-19
Fix for intermittent hex flashing failures due to address gaps

brianesquilona:


JanneKiiskila:
Commit description desperately needed explaining what's fixed, why and how the fix actually fixes the issue.

brianesquilona:
@JanneKiiskila 
DAPLink has a bug on handling address jumps on hex file that causes side effects of programming sectors without erasing it.

Issue #501, opened 2018-09-18
LPC546XX Fails to program above 256K


maclobdell:
Board: LPC54608
DAPLink Version: 0249

Flashing a binary < 256 KBytes works fine.  

Flashing a binary > 256 KBytes results in all 0's at 256 KBytes (0x40000) and above.  So applications larger than 256K will crash because instructions will not be present in the upper half as expected.  

Looks like the configuration is correct for 512K flash size.  I suspect a problem with the flash algorithm itself.

maclobdell:
@mmahadevan108

maclobdell:
cc @ashok-rao  This likely affects FF_LPC546XX also.

flit:
Can this ticket be closed now that #504 has been merged?

PR #504, opened 2018-09-21
Update LPC546xx Flash Algo

mmahadevan108:
Fix the erase chip function, it was not erasing the whole flash. This should fix Issue https://github.com/ARMmbed/DAPLink/issues/501


mmahadevan108:
cc @flit

PR #508, opened 2018-09-26
RapidIoT: Need separate board ID for the K64F and KW41Z

mmahadevan108:
Sharing 1 board ID does not work when running mbed-os

### DIFF
--- a/source/target/freescale/rapid_iot/target_reset.c
+++ b/source/target/freescale/rapid_iot/target_reset.c
@@ -38,6 +38,8 @@
 #define KW41         1
 
 static uint32_t mdm_id;
+extern char *board_id;
+char *board_id_kw41z = "0234";
 
 void target_before_init_debug(void)
 {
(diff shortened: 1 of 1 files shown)


mmahadevan108:
cc @flit @maclobdell

mmahadevan108:
Anything needed for this PR?

source/target/freescale/rapid_iot/target_reset.c
```diff
@@ -38,6 +38,8 @@
 #define KW41         1
 
 static uint32_t mdm_id;
+extern char *board_id;
+char *board_id_kw41z = "0234";
```
brianesquilona: Should we put all board IDs in the board source file instead; please see mkit_dk_dongle_nf5x.c.

mmahadevan108: This file is shared by the hexiwear board as well. Moving this to rapidiot board file will break the build for hexiwear.

flit: The problem is that the logic to detect the connected target is in this `targe_reset.c` file. If we moved the detect logic to the board source file, it would have to be duplicated for RapidIoT and Hexiwear. So no perfect solution here.

mmahadevan108: @brianesquilona. As Chris mentioned this would cause code duplication. Can we stick with the current implementation?

brianesquilona: @mmahadevan108 , This PR will be merged in a bit

PR #510, opened 2018-10-03
Added non-sequential non-increasing address support

### LINKED ISSUE #506: Silent failure in drag&drop flashing with non-sequential addresses
When flashing via drag and drop of a hex file, DAPLink needs the addresses to be sequential: 

https://github.com/ARMmbed/DAPLink/blob/ab7abed209eb166a98a8ecc69b05a4dcbe86adc6/source/daplink/drag-n-drop/flash_manager.c#L127-L134

If we have a file that first tries to write to address `0x1000_10C0` and then to `0x0000_0000` (example file [uicr-micropython-pycode.hex.txt](https://github.com/ARMmbed/DAPLink/files/2420961/uicr-micropython-pycode.hex.txt)) we get an clear error in the form of a fail.txt and assert.txt files:

```
error: An internal error has occurred
type: internal
```

```
Assert
File: ..\..\..\source\daplink\drag-n-drop\flash_manager.c
Line: 131
Source: Application
```

This is fine, but if we first start to write from address `0x0000_0000`, then jump to address `0x1000_10C0` and then jump back to `0x0003_E000` (example file [micropython-uicr-pycode.hex.txt](https://github.com/ARMmbed/DAPLink/files/2420966/micropython-uicr-pycode.hex.txt)), then the flashing fails silently.

The block starting at `0x0000_0000` and `0x1000_10C0` are flashed correctly, but the block starting at `0x0003_E000` is not flashed, which is expected, but there is no fail.txt or asset.txt files present.

Tested with DAPLink 0243 and 0250, in a micro:bit.

A test on DAPLink 0241 produces an assert.txt file but no fail.txt.

Although I understand why the sequential addresses enforcement has been included, it is worth noting that in contrast DAPLink 0234 seems to cope with this specific example correctly and flashes all areas as expected without showing any errors.

brianesquilona:
Removed address > flash_end as end of flashing. Non-increasing address will cause a sector setup, page erase if supported.

closes #506

### DIFF
--- a/source/daplink/drag-n-drop/flash_decoder.c
+++ b/source/daplink/drag-n-drop/flash_decoder.c
@@ -341,16 +341,7 @@ static bool flash_decoder_is_at_end(uint32_t addr, const uint8_t *data, uint32_t
         case FLASH_DECODER_TYPE_INTERFACE:
             end_addr = DAPLINK_ROM_IF_START + DAPLINK_ROM_IF_SIZE;
             break;
-
-        case FLASH_DECODER_TYPE_TARGET:
-            if (g_board_info.target_cfg) {
-                end_addr = g_board_info.target_cfg->flash_end;
-            }
-            else {
-                return false;
-            }
-            break;
-
+            
         default:
             return false;
     }
--- a/source/daplink/drag-n-drop/flash_decoder.c
+++ b/source/daplink/drag-n-drop/flash_decoder.c
@@ -56,6 +56,7 @@ static uint32_t initial_addr;
 static uint32_t current_addr;
 static bool flash_initialized;
 static bool initial_addr_set;
+static bool flash_type_target_bin;
 
 static bool flash_decoder_is_at_end(uint32_t addr, const uint8_t *data, uint32_t size);
 
@@ -65,7 +66,9 @@ flash_decoder_type_t flash_decoder_detect_type(const uint8_t *data, uint32_t siz
     util_assert(size >= FLASH_DECODER_MIN_SIZE);
     // Check if this is a daplink image
     memcpy(&info, data + DAPLINK_INFO_OFFSET, sizeof(info));
-
+    if(!addr_valid){ //reset until we know the binary type
+        flash_type_target_bin = false;
+    }
     if (DAPLINK_HIC_ID == info.hic_id) {
         if (DAPLINK_BUILD_KEY_IF == info.build_key) {
             // Interface update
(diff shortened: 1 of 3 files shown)


microbit-carlos:
I've tested a build from this PR (commit 06aea3f6ef597b04372de8d1cbd1b77576cd1aae) and created a few test hex files. This one in particular always fails with the following fail.txt error:
```
error: The transfer timed out.
type: transient, user
```
[flash-uicr-non-sequential.hex.zip](https://github.com/ARMmbed/DAPLink/files/2698902/flash-uicr-non-sequential.hex.zip)

microbit-carlos:
Btw, is DAPLink designed to write data to RAM?
A hex file that writes to all flash, then all user UICR, and then to all RAM causes this issue on macOS:

image

And then the micro:bit drive disappears until the micro:bit is disconnected and connected again.

[full-flash-full-uicr-full-ram.hex.zip](https://github.com/ARMmbed/DAPLink/files/2698919/full-flash-full-uicr-full-ram.hex.zip)

flit:
@microbit-carlos No, hex files must not include data written to RAM. All data is passed to the flash algo, which typically will error out if an address is not within flash. (You *could* have a flash algo that copied data to RAM if the address was in a RAM region, but afaik no flash algos support this.)

microbit-carlos:
Thanks for the clarification @flit, that makes a lot of sense.

Slightly related, but a bit of a side question. MakeCode currently stores metadata in the hex file by placing it in RAM after an end of file record. This seems to be working fine (in the sense that the flash and UICR data is written correctly) with DAPLink 0241+, but causes some issues on 0234, so it is likely to be changed. Nevertheless any inside in this issue would be very welcomed: https://github.com/Microsoft/pxt-microbit/issues/1631

brianesquilona:
@microbit-carlos  I have seen 2 issues from the first hex file. 
1. Hex decoding cannot decode the last line because there are no line feed.
2. Current non-sequential support cannot handle decreasing address on same write block or sector.

I have attached a bin if you can test it again before I update this PR.
[kl26z_microbit_if_crc.zip](https://github.com/ARMmbed/DAPLink/files/2715410/kl26z_microbit_if_crc.zip)

microbit-carlos:
Thanks @brianesquilona, I've fixed the script that generated that hex to add a new line at the end and the file doesn't time out anymore.

Are you updating the PR to merge it? Or still working on the feature?

brianesquilona:
@microbit-carlos , if you can test the bin I have attached in my previous comment, it would be great, this bin does not depend on the new line. Can you also confirm all your hex files to be working with this bin? Except for the one writing in RAM, this will cause a fault.

I will update this pr if this bin works out fine, the reason is that if it fails on any of your hex files, I would want the serial logs. Thanks

microbit-carlos:
@brianesquilona, sorry when I realised the bin file contained updates I updated my last comment instead of adding a new one.

Yep, I tested the bin that one works with and without the new line at the end of the file.
This PR also worked with all the test files we had from our editors and generated for testing (except for the ones writtng to RAM, as you pointed out).

PR #513, opened 2018-10-11
Added DAPABORT logic when initialization fails

brianesquilona:
Cause by reset, or download to target abort

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -156,22 +156,17 @@ uint8_t swd_write_dp(uint8_t adr, uint32_t val)
     uint8_t data[4];
     uint8_t ack;
 
-    switch (adr) {
-        case DP_SELECT:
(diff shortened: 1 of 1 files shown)


source/daplink/interface/swd_host.c
```diff
@@ -156,22 +156,16 @@ uint8_t swd_write_dp(uint8_t adr, uint32_t val)
...
-            }
-
-            dap_state.select = val;
-            break;
-
-        default:
-            break;
-    }
+    //check if the right bank is already selected
+    if(adr == DP_SELECT && dap_state.select == val)
```
flit: Please check formatting on these changes: Spaces between **if** and other keywords and parentheses, and between parentheses and braces. No spaces on inside of parentheses, eg `(foo)` not `( foo )`. Always use braces even for single statements (like line 161 here). Sorry to be pedantic 😁

Issue #523, opened 2018-11-02
DAPlink build fails on linux


studavekar:
List of failures

1) `  #include "system_sam3u.h` , ` #include "lpc43xx.h"`
```
Building project DAPLink (sam3u2c_bl, ARM)
Scan: DAPLink
Compile [ 17.6%]: main.c
[Error] sam3u2c.h@219,27:  #5: cannot open source input file "system_sam3u.h": No such file or directory
[ERROR] "./source/hic_hal/atmel/sam3u2c/sam3u2c.h", line 219: Error:  #5: cannot open source input file "system_sam3u.h": No such file or directory
  #include "system_sam3u.h"
                           ^
./source/daplink/bootloader/main.c: 0 warnings, 1 error

[mbed] ERROR: "/home/arm/daplink-pyenv/bin/python" returned error.
       Code: 1
       Path: "/home/arm/repo/DAPLink"
       Command: "/home/arm/daplink-pyenv/bin/python -u /home/arm/repo/DAPLink/mbed-os/tools/make.py -t ARM -m sam3u2c_bl --profile custom_profile.json --source . --build ./BUILD/SAM3U2C_BL/ARM-CUSTOM_PROFILE"
       Tip: You could retry the last command with "-v" flag for verbose output
---
Error - mbed compile!
```

2) `NameError: global name 'WindowsError' is not defined`
```
Traceback (most recent call last):
  File "tools/mbedcli_compile.py", line 92, in <module>
    main()
  File "tools/mbedcli_compile.py", line 65, in main
    generate_version_file(version_git_dir)
  File "/home/arm/repo/DAPLink/tools/pre_build_script.py", line 75, in generate_version_file
    except (CalledProcessError, WindowsError):
NameError: global name 'WindowsError' is not defined
```

3) Fails to create a release because of upper case PATH, it works on windows as its case-insensitive

cederom:
Does this mean DAPLink build on ARMCC/GCC is making its first steps? :-)

--
CeDeROM, SQ7MHZ, http://www.tomek.cedro.info

PR #524, opened 2018-11-02
fix linux build errors

studavekar:
This should address issue https://github.com/ARMmbed/DAPLink/issues/523


brianesquilona:
@studavekar Thanks! Coming up, linux build ci :)

Issue #516, opened 2018-10-24
RZ_A1H Flash algorithm write command FAILURE seen when automation turned on


studavekar:
DETAILS.TXT 
```
 DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 5500000003e0fb7200000000000000000000000097969902
HIC ID: 97969902
Auto Reset: 0
Automation allowed: 1
Overflow detection: 0
Daplink Mode: Interface
Interface Version: 0250
Git SHA: 6fece7592b9f0a57c032e48bfb70bd53d02dc33c
Local Mods: 0
USB Interfaces: MSD, CDC, HID, WebUSB
Interface CRC: 0x74181e66
Remount count: 1
URL: https://mbed.org/device/?code=5500000003e0fb7200000000000000000000000097969902?version=0250?target_id=@T
```

FAIL.TXT
```
error: Flash algorithm write command FAILURE
type: target
```

Steps to reproduce 
1 ) Turn on the automation mode
2) copy a fimrware binary on device. It fails to program.

toyowata:
cc @ARMmbed/team-renesas @tomoyamanaka

TomoYamanaka:
We are under the investigation of the causes. It seems that the read error occurs at validation processing after writing.

TomoYamanaka:
We addressed this issue with PR #525, so could you check after it was merged?

studavekar:
Tested it locally fixes the issue.

PR #525, opened 2018-11-05
Fixed write command when automation turned on (GR-PEACH, GR-LYCHEE)

d-kato:
Fixed the issues #516


studavekar:
Thanks @d-kato for fix.

Issue #522, opened 2018-11-02
RZ_A1H fails to program with "The interface firmware FAILED to reset/halt the target MCU"


studavekar:
DAPLINK details:
```
DETAILS.TXT

# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 550000000f01006100000000000000000000000097969902
HIC ID: 97969902
Auto Reset: 0
Automation allowed: 1
Overflow detection: 1
Daplink Mode: Interface
Interface Version: 0250
Git SHA: 682d8303e37355532402b8d93c4f240a3cec02a9
Local Mods: 0
USB Interfaces: MSD, CDC, HID, WebUSB
Interface CRC: 0x74181e66
Remount count: 136
URL: https://mbed.org/device/?code=550000000f01006100000000000000000000000097969902?version=0250?target_id=@T
```

FAIL.TXT
```
error: The interface firmware FAILED to reset/halt the target MCU
type: target
```
Steps to reproduce same as https://github.com/ARMmbed/DAPLink/issues/512 

Can reproduce above failure even with https://github.com/ARMmbed/DAPLink/pull/513 pulled in build. This target is Cortex-A might need a different fix.

brianesquilona:
#528

PR #526, opened 2018-11-05
Added DAPABORT logic when initialization fails for Cortex-A

d-kato:
Fixed the issues #522


brianesquilona:
@d-kato, I'm afraid I cannot merge this PR.

I have noticed some problems with RZ_A1H as target.
1. Writing the same value to DP_SELECT won't let me find the target debug registers (implementation dependent state?).
2. DAPABORT will do the same.

Instead we can prevent the issue through this PR #528

PR #528, opened 2018-11-05
Added checking to prevent target reset while flashing

brianesquilona:
It is implementation dependent that some platforms recovers DP and AP from DAPABORT, some wont. i.e. Cortex-A targets


studavekar:
@d-kato can you please review. Local testing looks ok!

d-kato:
LGTM 👍

brianesquilona:
daplink_build_mbedcli_pr 40
daplink_test_mbedcli_pr 28

PR #544, opened 2018-11-27
Add JTAG support into LPC11U35

### LINKED ISSUE #391: DAPLink, DAP_JTAG, LPC11U35
I am trying to include in the DAPLink Hw, the JTAG option. I define DAP_JTAG, the SWD work ok in Keil,  JTAG mode doesn't work. 
I found some bugs:
// TDI Pin                                      PIO0_17
#define PIN_TDI_PORT                 17        //I changed by 0
#define PIN_TDI_BIT                     17
// SWO/TDO Pin                           PIO0_9
#define PIN_TDO_PORT                9        //I changed by 0
#define PIN_TDO_BIT                    9
However, after these fixed still doesn't work. Any ideas?

kkitayam:
~~This PR consists of two commits.~~
4815db664d2e8e013e5800954888e9aefec5746e fixes #391.
~~e8c0e0c0840376b0a85546d050ccf0cd070c09e6 adds JTAG support for drag-n-drop programming.~~

## Fix #391

`PORT_JTAG_SETUP()` has no settings for `PIN_SWCLK_PORT` and `PIN_SWDIO_PORT` pins.
So JTAG had not been working correctly. This patch fixes this issue.

To enable JTAG, `DAP_JTAG=1` needs to be added to `records/board/your_board.yml`.

```YAML
common:
    macros:
        - DAP_JTAG=1
```

## ~~Add JTAG support for drag-n-drop programming~~

~~Functions for programming flash memory call CMSIS-DAP subroutines, which controls SWD directly.
So DAPLink cannot program via JTAG in spite of the above patch.
We implement the same features via JTAG at `daplink/interface/jtag_host.{c,h}`.
And wrapper functions are added to choose an interface (SWD or JTAG) at `daplink/interface/dap_host.{c,h}`.~~

~~To enable this patch, `source/target/<mfg>/target_reset.c` needs to be modified.
The prefix `swd_` should be changed to the prefix `dap_`.
It is shown below.~~

```C
#include "target_reset.h"
#include "dap_host.h"
void target_before_init_debug(void) {}
uint8_t target_unlock_sequence(void) {}
uint8_t target_set_state(TARGET_RESET_STATE state)
{
    return dap_set_target_state_hw(state);
}
uint8_t security_bits_set(uint32_t addr, uint8_t *data, uint32_t size)
{
    return 0;
}
```

~~If the target device is Cortex-A, `DEBUG_REGSITER_BASE_OVERRIDE` needs to be added to `records/board/your_board.yml`
It is shown below.~~

```YAML
common:
    macros:
        - CONF_JTAG
        - TARGET_MCU_CORTEX_A
        - DEBUG_REGSITER_BASE_OVERRIDE=0x80040000
```

~~These patches were tested on bluninja(lpc11u35_blueninja_if) board for Cortex-M target.
For Cortex-A target, we used the other board.~~

### DIFF
--- a/source/hic_hal/nxp/lpc11u35/DAP_config.h
+++ b/source/hic_hal/nxp/lpc11u35/DAP_config.h
@@ -42,12 +42,7 @@ Provides definitions about:
 // Configure JTAG option
 #if defined(BOARD_BAMBINO_210) || defined(BOARD_BAMBINO_210E)
 // LPC43xx multicore targets require JTAG to debug slave cores
-#define CONF_JTAG
-#endif
-
-// Configure SWD option by default
-#if !defined(CONF_JTAG) && !defined(CONF_SWD)
-#define CONF_SWD
+#define DAP_JTAG 1
 #endif
 
 /// Processor Clock of the Cortex-M MCU used in the Debug Unit.
@@ -64,18 +59,14 @@ Provides definitions about:
 
 /// Indicate that Serial Wire Debug (SWD) communication mode is available at the Debug Access Port.
 /// This information is returned by the command \ref DAP_Info as part of <b>Capabilities</b>.
-#if defined(CONF_SWD)
+#ifndef DAP_SWD
 #define DAP_SWD                 1               ///< SWD Mode:  1 = available, 0 = not available
-#else
-#define DAP_SWD                 0               ///< SWD Mode:  1 = available, 0 = not available
 #endif
 
 /// Indicate that JTAG communication mode is available at the Debug Port.
 /// This information is returned by the command \ref DAP_Info as part of <b>Capabilities</b>.
-#if defined(CONF_JTAG)
-#define DAP_JTAG                1               ///< JTAG Mode: 1 = available
-#else
-#define DAP_JTAG                0               ///< JTAG Mode: 0 = not available
+#ifndef DAP_JTAG
+#define DAP_JTAG                0               ///< JTAG Mode: 1 = available, 0 = not available
 #endif
 
 /// Configure maximum number of JTAG devices on the scan chain connected to the Debug Access Port.
@@ -84,10 +75,12 @@ Provides definitions about:
 
 /// Default communication mode on the Debug Access Port.
 /// Used for the command \ref DAP_Connect when Port Default mode is selected.
-#if (DAP_SWD != 0)
+#if (DAP_SWD == 1)
 #define DAP_DEFAULT_PORT        1               ///< Default JTAG/SWJ Port Mode: 1 = SWD, 2 = JTAG.
-#else
+#elif (DAP_JTAG == 1)
 #define DAP_DEFAULT_PORT        2               ///< Default JTAG/SWJ Port Mode: 1 = SWD, 2 = JTAG.
+#else 
+#error Must enable DAP_SWD and/or DAP_JTAG
 #endif
 
 /// Default communication speed on the Debug Access Port for SWD and JTAG mode.
@@ -189,15 +182,23 @@ __STATIC_INLINE void PORT_JTAG_SETUP(void)
     LPC_GPIO->SET[PIN_SWCLK_PORT] = PIN_SWCLK;
     LPC_GPIO->SET[PIN_SWDIO_PORT] = PIN_SWDIO;
 #if !defined(PIN_nRESET_FET_DRIVE)
-    // open drain logic
+    /* Open drain logic (for ordinary board).
+     * nRESET line should be pulled up by the board.
+     * To output high level (reset release signal),
+     * set as input direction. */
     LPC_GPIO->DIR[PIN_nRESET_PORT] &= ~PIN_nRESET;
     LPC_GPIO->CLR[PIN_nRESET_PORT] = PIN_nRESET;
 #else
-    // FET drive logic
+    /* FET drive logic (for special board like as blueninja)
+     * This setting treats nRESET line as positive logic.
+     * High level indicates that reset signal is asserted.
+     * Low level indicates that reset signal is deasserted. */
     LPC_GPIO->DIR[PIN_nRESET_PORT] |= PIN_nRESET;
     LPC_GPIO->CLR[PIN_nRESET_PORT] = PIN_nRESET;
 #endif
+    // SWCLK and TCK are aliases for the same line.
     LPC_GPIO->DIR[PIN_SWCLK_PORT] |= PIN_SWCLK;
+    // SWDIO and TMS are aliases for the same line.
     LPC_GPIO->DIR[PIN_SWDIO_PORT] |= PIN_SWDIO;
 
     LPC_GPIO->SET[PIN_TDI_PORT]  =  PIN_TDI;
@@ -420,8 +421,13 @@ __STATIC_FORCEINLINE void     PIN_nTRST_OUT(uint32_t bit)
 __STATIC_FORCEINLINE uint32_t PIN_nRESET_IN(void)
 {
 #if !defined(PIN_nRESET_FET_DRIVE)
+    // open drain logic
     return LPC_GPIO->B[PIN_nRESET_BIT + PIN_nRESET_PORT * 32] & 0x1;
 #else
+    /* FET drive logic (for special board like as blueninja)
+     * This setting treats nRESET line as positive logic.
+     * High level indicates that reset signal is asserted.
+     * Low level indicates that reset signal is deasserted. */
     return (LPC_GPIO->B[PIN_nRESET_BIT + PIN_nRESET_PORT * 32] & 0x1) == 1 ? 0 : 1;
 #endif
 }
--- a/source/hic_hal/nxp/lpc11u35/DAP_config.h
+++ b/source/hic_hal/nxp/lpc11u35/DAP_config.h
@@ -45,6 +45,11 @@ Provides definitions about:
 #define CONF_JTAG
 #endif
 
+// Configure SWD option by default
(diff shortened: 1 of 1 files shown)


kkitayam:
We have rebased onto the head of master. After that, we tested it on the same boards.

jxltom:
What's the status of this PR? Does this mean the daplink based on lpc11u35 doesn't support JTAG actually?

kkitayam:
@jxltom 

> What's the status of this PR?

It needs rebase to head of main branch.

> Does this mean the daplink based on lpc11u35 doesn't support JTAG actually?

The main branch works with SWD interface only. It does not work with JTAG interface. Because some pins settings are wrong.

jxltom:
@kkitayam, thanks for your work, and clarification

source/daplink/interface/dap_host.c
```diff
@@ -0,0 +1,150 @@
...
+    return swd_set_target_state_sw(state);
+#else
+    return jtag_set_target_state_sw(state);
+#endif
+}
+
+void dap_set_reset_connect(DAP_CONNECT_TYPE type)
+{
+#if (DAP_SWD == 1)
+    return dap_set_reset_connect(type);
```
mbrossard, line 139: This looks like it would create an infinite loop

kkitayam: Thank you for your comment. I will fix it.

mbrossard:
I don't know if @flit or @brianesquilona will agree (wait until they weigh in) but given the symmetry `source/daplink/interface/dap_host.c` I would suggest removing it by `#define` in `source/daplink/interface/dap_host.h` or add `__INLINE`.

I would also suggest either moving the JTAG parts from `dap_host.h` into a separate `jtag_host.h`, or the content of `swd_host.h` into `dap_host.h`.

mathias-arm:
I have looked again at this PR, and I have several comments:
- I think the changes in `source/hic_hal/nxp/lpc11u35/DAP_config.h` should be separated from the rest. It would be nice to have an explanation about the added lines regarding SWCLK/SWDIO. There should be some comment in the code about `PIN_nRESET_FET_DRIVE`.
- It does not make sense to me to have SWD and JTAG be mutually exclusive. Not many LPC11U35 allow JTAG operations, they must be even fewer that only work for JTAG. So like your solution I think the default should be SWD on/JTAG off. If  DAP_JTAG=1  is defined, then both are off. If DAP_JTAG=1 and DAP_SWD=0 are defined then only JTAG is enabled. `DAP_Config.h` should look something like this (the `CONF_JTAG` should be removed).
  ```c
  #ifndef DAP_SWD
  #define DAP_SWD                 1
  #endif

  #ifndef DAP_JTAG
  #define DAP_JTAG                0
  #endif

  #if (DAP_SWD == 1)
  #define DAP_DEFAULT_PORT        1
  #elif (DAP_JTAG == 1)
  #define DAP_DEFAULT_PORT        2
  #else 
  #error "Must enable DAP_SWD and/or DAP_JTAG"
  #endif```
- Regarding the JTAG support for drag&drop.  This should be done in a separate pull-request. `dap_host.c` should be replaced with defines in `dap_host.h`:
  ```c
  #if (DAP_SWD == 1)
  #define dap_init swd_init
  ...
  #elif (DAP_JTAG == 1)
  #define dap_init jtag_init
  ...
  #else 
  #error "Must enable DAP_SWD and/or DAP_JTAG"
  #endif
  ```
  The JTAG parts of `dap_host.h` should be moved into a separate `jtag_host.h`.

kkitayam:
Thank you for your comments. I am happy to get your comments regarding this PR.

Let me go back to talk with our team regarding your comments. After getting the agreement of our team, I will separate this PR to support for JTAG and to support for drag and drop via JTAG. BTW, it is hard to test this PR with our boards because of I am currently working from home, my boss does not allow me to bring our boards to home. Although no tests run on our boards I will separate to two PRs.

To separate to two PRs, should I remove this PR and crate new two PRs? Or, should I keep this PR for only to support for JTAG and crate a additional PR to support for drag and drop via JTAG?

mathias-arm:
This PR has the right title for keeping the LPC11U35-specific part.

@stephendpmurphy and @elfmimi have been talking in this [discussion](https://github.com/ARMmbed/DAPLink/discussions/770#discussioncomment-817832) about testing JTAG. If we can find a couple of options with a probe (LPC11U35 or not) and a target that work with JTAG, we might be able to help with testing.

kkitayam:
OK, I will move support for drag and drop from this PR to another one. After that is done, could I ask you for advise again regarding test.

kkitayam:
I've finished removing commits regarding D&D and adding comments for `PIN_nRESET_FET_DRIVE`. I checked only that there are no build errors. If you don't mind, could you check if it works?

mathias-arm:
I tried getting those patches applied to `develop` with Raspberry Pi 2, but it looks like I will need to take out my soldering iron because the IBDAP, that I [mentioned a few days ago](https://github.com/ARMmbed/DAPLink/discussions/770#discussioncomment-848967) does not have the right pin mapping.

mathias-arm:
If @elfmimi and/or @stephendpmurphy could share the result of their tests on `develop` with this patch that should be enough validation.

elfmimi:
I prepared it like this.
```
git fetch origin develop
git fetch origin pull/544/head:pull/544
git checkout pull/544
git rebase --onto origin/develop HEAD~2 HEAD
```
I had to add - DAP_JTAG=1 to records/hic_hal/lpc11u35.yaml . then build it like this.
> \> progen generate -t make_gcc_arm -p lpc11u35_if -b

Then test it like this.
> \> openocd -f interface/cmsis-dap.cfg -c "transport select jtag" -f target/bcm2835.cfg

I could successfully discover RaspberryPi Zero (BCM2835) in JTAG chain.

```
Info : JTAG tap: bcm2835.cpu tap/device found: 0x07b7617f (mfg: 0x0bf (Broadcom), part: 0x7b76, ver: 0x0)
Info : found ARM1176
Info : bcm2835.cpu0: hardware has 6 breakpoints, 2 watchpoints
Info : starting gdb server for bcm2835.cpu0 on 3333
Info : Listening on port 3333 for gdb connections
```

stephendpmurphy:
I've yet to get around to testing this, but @elfmimi results look promising :+1:

mbrossard:
I found a board that not only has JTAG pins on the Cortex-M connector but also a configuration included in openocd 0.11.0: the `GR PEACH` (ironically the board has one LPC11U35 HIC on-board already, but the TDI signal is not connected).

```$ openocd -f interface/cmsis-dap.cfg -c "transport select jtag" -c "adapter speed 1000" -f board/renesas_gr_peach.cfg
[...]
Info : Using CMSIS-DAPv2 interface with VID:PID=0x0d28:0x0204, serial=00000080077ca99d00000000000000000000000097969902
Info : CMSIS-DAP: SWD  Supported
Info : CMSIS-DAP: JTAG Supported
Info : CMSIS-DAP: FW Version = 2.0.0
Info : CMSIS-DAP: Serial# = 00000080077ca99d00000000000000000000000097969902
Info : CMSIS-DAP: Interface Initialised (JTAG)
Info : SWCLK/TCK = 1 SWDIO/TMS = 1 TDI = 1 TDO = 1 nTRST = 0 nRESET = 1
Info : CMSIS-DAP: Interface ready
Info : clock speed 1000 kHz
Info : cmsis-dap JTAG TLR_RESET
Info : cmsis-dap JTAG TLR_RESET
Info : JTAG tap: r7s72100.cpu tap/device found: 0x4ba00477 (mfg: 0x23b (ARM Ltd), part: 0xba00, ver: 0x4)
Info : r7s72100.ca9: hardware has 6 breakpoints, 4 watchpoints
Info : starting gdb server for r7s72100.ca9 on 3333```
```

It seems to be working!

@kkitayam: do you want to include a patch to enable JTAG for the `blueninja`?

mbrossard (approved):

kkitayam:
@mbrossard @elfmimi @stephendpmurphy 
Thank you very much for your help. I am happy to confirm that the patch works.

> do you want to include a patch to enable JTAG for the blueninja?

No, it isn't needed.
At the time, I needed a JTAG interface for our own internal projects which uses another board. I had used the blueninja to check if I create any bugs into the existing code.

kkitayam:
@mbrossard @elfmimi @stephendpmurphy
Thank you very much for your help. I am happy to confirm that the patch works.

> do you want to include a patch to enable JTAG for the blueninja?

If you don't mind, could you include the codes wrapped with `PIN_nRESET_FET_DRIVE` as it is.
The part of the codes wrapping with `PIN_nRESET_FET_DRIVE` of `PORT_JTAG_SETUP` function is that I had needed for our internal project. If you decide the part is not needed, I will remove the part.
The part of the codes wrapping with `PIN_nRESET_FET_DRIVE` of `PIN_nRESET_IN` function is a bug of blueninja.

mathias-arm:
I meant updating the `blueninja.yaml`, for example [9d6ce1e](https://github.com/mathias-arm/DAPLink/commit/9d6ce1e0f00d5d89d6f43ffe72f1ae47384b143e).

kkitayam:
Thanks for the clarification. No, `blueninja.yaml` doesn't need updating.
`PIN_nRESET_FET_DRIVE` is already defined at [here](https://github.com/ARMmbed/DAPLink/blob/main/source/board/override_blueninja/IO_Config_Override.h#L33). And, it is no problem that the bluninja can't use JTAG.

mathias-arm:
Ok. I did not know it was already defined.

PR #551, opened 2018-12-04
Fix persistent storage to actual files created

brianesquilona:
Added size check, large size flashing in mac causing size revert.


brianesquilona:
daplink_test_mbedcli_pr 48

PR #553, opened 2018-12-04
Added multiple ram and flash region check

### LINKED ISSUE #549: Add multiple region support for ram and flash

brianesquilona:
closes #549

### DIFF
--- a/source/hic_hal/target_config.h
+++ b/source/hic_hal/target_config.h
@@ -40,6 +40,15 @@ extern "C" {
 // This can vary from target to target and should be in the structure or flash blob
 #define TARGET_AUTO_INCREMENT_PAGE_SIZE    (1024)
 
+//Additional flash and ram regions
+#define MAX_EXTRA_FLASH_REGION                3
+#define MAX_EXTRA_RAM_REGION                  3
+
+typedef struct region_info {
+    uint32_t start;
+    uint32_t end;
+} region_info_t;
+
 /**
  @struct target_cfg_t
  @brief  The firmware configuration struct has unique about the chip its running on.
@@ -55,6 +64,8 @@ typedef struct target_cfg {
     uint8_t erase_reset;            /*!< Reset after performing an erase */
     const sector_info_t* sectors_info; 
     int sector_info_length;
+    region_info_t extra_flash[MAX_EXTRA_FLASH_REGION + 1]; //!< Extra flash regions.
+    region_info_t extra_ram[MAX_EXTRA_RAM_REGION + 1]; //!< Extra RAM regions.
 } target_cfg_t;
 
 extern target_cfg_t target_device;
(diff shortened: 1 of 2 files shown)


brianesquilona:
Closes #549

source/hic_hal/target_config.h
```diff
@@ -40,6 +40,9 @@ extern "C" {
 // This can vary from target to target and should be in the structure or flash blob
 #define TARGET_AUTO_INCREMENT_PAGE_SIZE    (1024)
 
+//Additional flash and ram regions
+#define MAX_ADD_FLASH_REGION                3
+#define MAX_ADD_RAM_REGION                  3
 /**
```
flit: Can you please insert a blank line here between the #define and start of the comment?

source/hic_hal/target_config.h
```diff
@@ -55,6 +58,14 @@ typedef struct target_cfg {
     uint8_t erase_reset;            /*!< Reset after performing an erase */
     const sector_info_t* sectors_info; 
     int sector_info_length;
+    
+    /*!< Additional flash and ram regions */
```
flit: I'd rather have the extra regions done as structs that pair together start and end. And instead of requiring count fields, treat an entry with start and end both set to 0 as the terminator. You might want to increase the counts so you still have max 3 regions + terminator.

Generally, using 0-value list terminators makes it easier to define lists in code because you don't have to make sure to keep a count value defined somewhere in sync. And in a case like this, it saves a little space since you're always allocating the same number of max extra regions.

Another idea: maybe use "extra" instead of "add"? Just so that you don't have to use an incomplete word.

Example:
```c
typedef struct region_info {
    uint32_t start;
    uint32_t end;
} region_info_t;

typedef struct target_cfg {
    // ... existing fields ...

    region_info_t extra_flash[MAX_EXTRA_FLASH_REGION]; //!< Extra flash regions.
    region_info_t extra_ram[MAX_EXTRA_RAM_REGION]; //!< Extra RAM regions.
} target_cfg_t;
```

Then a target definition would look like this:

```c
const target_cfg_t mytarget = {
        // ... standard stuff ...

        .extra_flash = {
                { 0x10000000, 0x1000FFFF },
        },
        .extra_ram = {
                { 0x30000000, 0x30007FFF },
        },
    };
```

(Btw, the `<` would have to be removed on this comment on line 62. A doxygen comment marker of `!<` indicates that the comment _follows_ the definition. If the comment precedes the definition, only `!` should be used.)

source/hic_hal/target_config.h
```diff
@@ -55,6 +58,14 @@ typedef struct target_cfg {
...
+    uint8_t add_flash_region;
+    uint8_t add_ram_region;
+    uint32_t add_flash_start[MAX_ADD_FLASH_REGION];
+    uint32_t add_flash_end[MAX_ADD_FLASH_REGION];
+    uint32_t add_ram_start[MAX_ADD_FLASH_REGION];
```
flit: Note that `MAX_ADD_FLASH_REGION` is used here on the additional ram regions.

PR #569, opened 2019-01-22
Edit USERS-GUIDE.md

AnotherButler:
Make initial copy edits to file.

### DIFF
--- a/docs/USERS-GUIDE.md
+++ b/docs/USERS-GUIDE.md
@@ -1,40 +1,41 @@
-# DAPLink Users Guide
-There are three interfaces that DAPLink provides. These are drag-n-drop programming, a serial port and debugging support. In addition DAPLink firmware can be updated using the drag-n-drop programming interface of the bootloader.
+# DAPLink users guide
 
+DAPLink provides three interface. These are drag-and-drop programming, a serial port and debugging support. In addition, you can update DAPLink firmware using the drag-and-drop programming interface of the bootloader.
 
-## Drag-n-drop Programming
-Program the target microcontroller by copying or saving a file in one of the supported formats to the DAPLink drive. Upon completion the drive will re-mount. If a failure occurs then the file ``FAIL.TXT`` will appear on the drive containing information about the failure.
+## Drag-and-drop programming
+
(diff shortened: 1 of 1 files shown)


docs/USERS-GUIDE.md
```diff
@@ -1,40 +1,41 @@
-# DAPLink Users Guide
-There are three interfaces that DAPLink provides. These are drag-n-drop programming, a serial port and debugging support. In addition DAPLink firmware can be updated using the drag-n-drop programming interface of the bootloader.
+# DAPLink users guide
 
+DAPLink provides three interface. These are drag-and-drop programming, a serial port and debugging support. In addition, you can update DAPLink firmware using the drag-and-drop programming interface of the bootloader.
```
loverdeg-ep: interfaces?

docs/USERS-GUIDE.md
```diff
@@ -1,40 +1,41 @@
-# DAPLink Users Guide
-There are three interfaces that DAPLink provides. These are drag-n-drop programming, a serial port and debugging support. In addition DAPLink firmware can be updated using the drag-n-drop programming interface of the bootloader.
+# DAPLink users guide
```
AnotherButler: @brianesquilona Could you please add an introduction about what DAPLink is and why you'd use it?

brianesquilona: @AnotherButler , can I redirect you to the readme file :) [The very first paragraph] (https://github.com/ARMmbed/DAPLink/blob/master/README.md)

AnotherButler: Aha. I was looking at the README in docs, not in the overall repo.

docs/USERS-GUIDE.md
```diff
@@ -1,40 +1,41 @@
...
 
-Note - Most DAPLink implementations support other baud rates in addition to the ones listed here.
+- 9600.
+- 14400.
+- 19200.
+- 28800.
+- 38400.
+- 56000.
+- 57600.
+- 115200.
```
flit: Can you please remove the periods after the baud rates. That is a list of items, not sentences. It looks really weird with the period. Same for supported file formats above.

flit: That's fine, and makes sense, for sentences. But these are not sentences here. I've *never* seen technical documentation that uses periods in cases like this.

PR #574, opened 2019-01-31
Change default rom settings auto_rst to 1.

brianesquilona:
Will trigger reset and run of the binary after flashing

### DIFF
--- a/source/daplink/settings/settings_rom.c
+++ b/source/daplink/settings/settings_rom.c
@@ -67,7 +67,7 @@ static cfg_setting_t config_rom_copy;
 
 // Configuration defaults in flash
 static const cfg_setting_t config_default = {
-    .auto_rst = 0,
+    .auto_rst = 1,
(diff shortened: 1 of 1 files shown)


flit:
I think we may want to set micro:bit to disable auto reset so its behaviour isn't changes.
cc @microbit-carlos @jaustin

microbit-carlos:
Thanks for checking @flit. Auto reset is enabled on the micro:bit so that the user code starts executing as soon as the file transfer is done.

The factory images ship with the interface settings/config data (the one at the last page of flash), which enables auto reset (and disables automation and overflow_detect), so changing the global default would still be okay for our builds.
Although if there were more config options added we would also want ensure we are configuring them for our factory builds, even if we set them to the same values as the defaults.

flit:
Thanks for the clarification! For some reason I thought that micro:bit had auto reset disabled, but I must have been thinking about automation.

PR #575, opened 2019-02-01
Rename reset_stop_on_reset() to reset_and_halt()

khariya:
Match the function name with pyocd tools.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 project_generator==0.9.13
 mbed_ls==1.6.2
 pyserial
-pyOCD==0.15.0
+pyOCD==0.16.0
 requests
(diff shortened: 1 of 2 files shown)


flit:
We can't merge this yet because DAPLink is pinned to pyOCD version 0.15.0, which doesn't have the `reset_and_halt()` name change. After v0.16.0 is released with the name change, we'll be able to merge this PR. But you'll need to update the pyOCD version in `requirements.txt`.

flit:
@khariya Can you update `requirements.txt` to set the pyOCD version to 0.16.0? Then we can merge in this PR. Thanks!

khariya:
I'm sorry for the delay.

PR #576, opened 2019-02-01
Add the missing include file

khariya:
Raised an error when compiling on uvision IDE v5.17 with ARM compiler v5.06.

### DIFF
--- a/source/hic_hal/gpio.h
+++ b/source/hic_hal/gpio.h
@@ -22,6 +22,7 @@
 #ifndef GPIO_H
 #define GPIO_H
 
+#include "stdbool.h"
 #include "IO_Config.h"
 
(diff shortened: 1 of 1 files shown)


source/hic_hal/gpio.h
```diff
@@ -22,6 +22,7 @@
 #ifndef GPIO_H
 #define GPIO_H
 
+#include "stdbool.h"
```
flit: `stdbool.h` should be included using angle brackets, i.e. `#include <stdbool.h>`, since it's a system header.

flit:
Interesting! I wonder why we don't see this error? Either way, it's good to have the fix.

flit:
You're fast! 😁

PR #583, opened 2019-02-06
Add MAX32620 HIC and MAX32625MBED board

khariya:


records/hic_hal/max32620.yaml
```diff
@@ -0,0 +1,18 @@
+common:
+    target:
+        - max32620
+    macros:
```
brianesquilona: Need to add a core here, for example:
`core:
        - Cortex-M4`

khariya: Do Cortex-M4 with and w/o FPU have same names in this file?
MAX32620 is based on Arm Cortex-M4 with FPU.

brianesquilona: Yes it should be M4F

khariya: Thanks for updating that!

PR #587, opened 2019-02-14
Added page erase and under reset connection to wise-1570

brianesquilona:


brianesquilona:
daplink_build_mbedcli_pr 111 daplink_test_mbedcli_pr 100

PR #596, opened 2019-03-01
Add HIC MAX32625PICO and MAX32620FTHR board

khariya:


test/info.py
```diff
@@ -163,10 +163,11 @@ def VENDOR_TO_FAMILY(x, y) : return (VENDOR_ID[x] <<8) | y
...
+    (   0x0262,     VENDOR_TO_FAMILY('NXP', 2),         'k20dx_frdmkl43z_if',                       'k20dx_bl',         'FRDM-KL43Z'                            ),
     (   0x0311,     VENDOR_TO_FAMILY('NXP', 1),         'k20dx_frdmk66f_if',                        'k20dx_bl',         'FRDM-K66F'                             ),
     (   0x0350,     VENDOR_TO_FAMILY('Stub', 1),        'k20dx_xdot_l151_if',                       'k20dx_bl',         None                                    ), # TODO - set target to 'MTS-xDot-L151CC' when mbed-os supports this
     (   0x0415,     VENDOR_TO_FAMILY('Stub', 1),        'max32620_max32625mbed_if',                 'max32620_bl',      'MAX32625'                              ),
+    (   0x0418,     VENDOR_TO_FAMILY('Stub', 1),        'max32625_max32620fthr_if',                 'max32625_bl',      'MAX32620'                              ),
     (   0x0450,     VENDOR_TO_FAMILY('Stub', 1),        'lpc11u35_mtb_ublox_odin_w2_if',            None,               None                                    ),
```
brianesquilona: Can we add an entry in the PROJECT_RELEASE_INFO like I did.
`('max32620_max32625mbed_if',                    False,      0x0000,     "bin"       ),`
This is to have a release binary for maxim every time there is a DAPLink release. You can also have edit this line for the max32620_max32625mbed_if if you want to support a legacy bootloader, but a DAPLink bootloader won't need it.

khariya: What do you mean by legacy bootloader?
Does the Offset field in PROJECT_RELEASE_INFO have to match the address in [daplink_addr.h](https://github.com/ARMmbed/DAPLink/blob/master/source/hic_hal/maxim/max32620/daplink_addr.h#L41)?

brianesquilona: DAPLink bootloader already follow's the map in the daplink_addr.h. This offset is for a different bootloader, in case you have one and supporting it. If the Legacy field is false, offset does not matter, but will be appended in the release name.

khariya: Got it. Thanks!
I'll add the max32620fthr to the list shortly.

brianesquilona:
@khariya I have tried the bootloader and interface on a MAX32625PICO, but both have enumeration issue on usb, should be easily replicated at your side.

PR #600, opened 2019-03-07
Dev feature no rtos

brianesquilona:
- Rtos type module in main yaml. 
- Bootloader projects with no rtos. 
- Interface are on rtx4
- No RTOS implementation and RTOS api files added
1. Added API for RTOS
2. Added No RTOS and RTX4 api implementation/port
3. Moved RTX_Config.c to rtos folder
4. Moved OS_CLOCK config to HICs
5. Moved OS_TICK config to RTOS

### DIFF
--- a/projects.yaml
+++ b/projects.yaml
@@ -20,6 +20,7 @@ module:
         - records/daplink/settings.yaml
         - records/daplink/settings_rom_stub.yaml
         - records/daplink/target_board.yaml
+        - records/rtos/rtos-none.yaml
     if: &module_if
         - *module_tools
         - records/usb/usb-core.yaml
@@ -68,7 +69,7 @@ module:
 projects:
     kl26z_bl:
         - *module_bl
-        - *module_hic_kl26z
+        - records/hic_hal/kl26z.yaml
         - records/board/kl26z_bl.yaml
     kl26z_if:
         - *module_if
@@ -76,11 +77,19 @@ projects:
         - records/family/all_family.yaml
     k20dx_bl:
         - *module_bl
-        - *module_hic_k20dx
+        - records/hic_hal/k20dx.yaml
         - records/board/k20dx_bl.yaml
     k20dx_if:
         - *module_if
         - *module_hic_k20dx
+        - records/family/all_family.yaml
+    k26f_bl:
+        - *module_bl
+        - records/hic_hal/k26f.yaml
+        - records/board/k26f_bl.yaml
+    k26f_if:
+        - *module_if
+        - *module_hic_k26f
         - records/family/all_family.yaml
     lpc11u35_if:
         - *module_if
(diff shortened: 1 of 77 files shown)


projects.yaml
```diff
@@ -36,542 +36,673 @@ module:
...
+    # RTOS types
+    rtx4_cm0: &module_rtx4_cm0
+        - records/rtos/rtos-rtx4-cm0.yaml
+    rtx4_cm3: &module_rtx4_cm3
+        - records/rtos/rtos-rtx4-cm3.yaml
+    rtx4_cm4: &module_rtx4_cm4
+        - records/rtos/rtos-rtx4-cm4.yaml
+    rtos_none: &module_rtos_none
+        - records/rtos/rtos-none.yaml
+
```
flit: Lets's simplify the changes here.
- Revert HIC yaml modules so they include the RTOS records.
- Remove RTOS yaml modules.
- For the bootloader projects, add the HIC yaml record plus RTOS-none yaml record directly.
- Add `records/rtos/rtos-none.yaml` directly into `module_bl`.

Example bootloader project:
```yaml
k20dx_bl:
  - *module_bl
  - records/hic_hal/k20dx.yaml
  - records/board/k20dx_bl.yaml
```

These changes will make it so the interface projects yaml don't have to change at all. All the differences are pushed onto the bootloader projects, which are few in number.

source/daplink/interface/main.c
```diff
@@ -107,30 +107,17 @@ static main_led_state_t msc_led_state = MAIN_LED_FLASH;
...
+void timer_task_30mS(void)
 {
-    uint8_t i = 0;
-    os_itv_set(3); // 30mS
-
-    while (1) {
-        os_itv_wait();
-        os_evt_set(FLAGS_MAIN_30MS, main_task_id);
-
-        if (!(i++ % 3)) {
```
flit: You could leave the logic in `timer_task_30mS()` for setting the 90 ms event (for both interface and bootloader). Just change `i` to be static. This would prevent you from having to move around the chunks of code in the main task.

source/rtos_none/SysTick_Handler.c
```diff
@@ -0,0 +1,101 @@
...
+static volatile rtosTick_t wait_counter = 0;
+
+static sysTickCb_t sysTickCb = NULL;
+static rtosTick_t tickFreq = 1;
+
+void sysTickInit(void)
+{
+  NVIC_ST_RELOAD  = OS_TRV;
+  NVIC_ST_CURRENT = 0;
+  NVIC_ST_CTRL    = 0x0007;
```
flit: Use the SysTick related register definitions and functions in the CMSIS-Core headers instead of writing your own. See [core_cm4.h:647](https://github.com/ARMmbed/DAPLink/blob/master/source/cmsis-core/core_cm4.h#L647) and [core_cm4.h:1695](https://github.com/ARMmbed/DAPLink/blob/master/source/cmsis-core/core_cm4.h#L1695).

source/rtos_none/rtos_api.h
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __RTOS_API_H__
+#define __RTOS_API_H__
```
flit: Should use the CMSIS-RTOS2 API header, but only implement those APIs which you need. This way you're sharing a single common API between no-rtos and rtos that won't get out of sync accidentally.

brianesquilona:
I did thought about this but rtx5 has its own copy in one of the include directory. Will revisit this again on switching to rtx5.

flit:
Ok, sounds good.

PR #603, opened 2019-03-08
K26F HIC and FRDM-K32W042 board

flit:
This PR adds a new NXP K26F HIC. It has high speed USB and SWO support enabled. The FRDM-K32W042 board is also added that uses this HIC.


tools/post_build_script.py
```diff
@@ -173,6 +173,10 @@ def post_build_script(input_file, output_file, board_id=None, family_id=None, bi
         offset_update.create_padded_image(output_file_legacy,
                                           output_file_legacy_5000,
                                           start, pad_addr, 0x40)
+    elif start == 0 or start == 0x80000 or start == 0x20000:
+        pass
+    else:
+        assert 0, "Unsupported start address 0x%x" % start
```
brianesquilona: lpc4322 has a start address of 0x1A010000, so the build script failed

flit: Ah, thanks for the explanation!

PR #604, opened 2019-03-08
Add MAX32630FTHR board

khariya:


source/family/maxim/max32630/target_reset.c
```diff
@@ -0,0 +1,55 @@
...
+
+uint8_t security_bits_set(uint32_t addr, uint8_t *data, uint32_t size)
+{
+    // if there are security bits in the programmable flash region
+    //  a check should be performed. This method is used when programming
+    //  by drag-n-drop and should refuse to program an image requesting
+    //  to set the device security. This can be performed with the debug channel
+    //  if needed.
+    return 0;
+}
```
brianesquilona: Can we delete this file? This is not really used.

khariya: Done!
I will submit a separate PR to remove equivalent file for other Maxim targets.

PR #612, opened 2019-03-20
Support for multiple flash algo.

### LINKED ISSUE #605: Support separate algos for each flash region
DAPLink supports extra flash memory regions in the `target_cfg_t` struct. However, it currently only supports a single flash algo, the `program_target_t` struct. This needs to be extended to support a separate flash algo for each memory region.

brianesquilona:
Note, boundary for addresses should be atleast sector size

Closes #605

### DIFF
--- a/source/daplink/interface/target_flash.c
+++ b/source/daplink/interface/target_flash.c
@@ -47,6 +47,7 @@ static error_t target_flash_erase_chip(void);
 static uint32_t target_flash_program_page_min_size(uint32_t addr);
 static uint32_t target_flash_erase_sector_size(uint32_t addr);
 static uint8_t target_flash_busy(void);
+static error_t target_flash_set(uint32_t addr);
 
 static const flash_intf_t flash_intf = {
     target_flash_init,
@@ -57,6 +58,7 @@ static const flash_intf_t flash_intf = {
     target_flash_program_page_min_size,
     target_flash_erase_sector_size,
     target_flash_busy,
+    target_flash_set,
 };
 
 static state_t state = STATE_CLOSED;
@@ -65,9 +67,35 @@ const flash_intf_t *const flash_intf_target = &flash_intf;
 
 static flash_func_t last_flash_func = FLASH_FUNC_NOP;
 
+//saved flash algo
+static program_target_t * current_flash_algo = NULL;
+
+//saved flash start from flash algo
+static uint32_t flash_start = 0;
+
+static program_target_t * get_flash_algo(uint32_t addr)
+{
+    flash_region_info_t * flash_region = g_board_info.target_cfg->extra_flash;
+
+    for (; flash_region->start != 0 && flash_region->end != 0; ++flash_region) {
+        if (addr >= flash_region->start && addr <= flash_region->end) {
+            flash_start = flash_region->start; //save the flash start
+            if (flash_region->flash_algo) {
+                return flash_region->flash_algo;
+            }else{
+                return NULL;
+            }
+        }
+    }
+    
+    //could not find a flash algo for the region; use default
+    flash_start = g_board_info.target_cfg->flash_start;
+    return g_board_info.target_cfg->flash_algo;
+}
+
 static error_t flash_func_start(flash_func_t func)
 {
-    const program_target_t *const flash = g_board_info.target_cfg->flash_algo;
+    program_target_t * flash = current_flash_algo;
 
     if (last_flash_func != func)
     {
@@ -79,7 +107,7 @@ static error_t flash_func_start(flash_func_t func)
 
         // Start a new function.
         if (FLASH_FUNC_NOP != func &&
-            0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->init, g_board_info.target_cfg->flash_start, 0, func, 0)) {
+            0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->init, flash_start, 0, func, 0)) {
             return ERROR_INIT;
         }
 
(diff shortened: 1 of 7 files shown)


source/daplink/interface/target_flash.c
```diff
@@ -253,16 +300,41 @@ static error_t target_flash_erase_chip(void)
...
-        status = flash_func_start(FLASH_FUNC_ERASE);
-        if (status != ERROR_SUCCESS) {
-            return status;
-        }
-        if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
-            return ERROR_ERASE_ALL;
+        program_target_t * flash, * start_flash_algo = last_flash_algo;
+        uint8_t index = 0;
+        
+        do{
```
flit: Space before brace.

source/daplink/interface/target_flash.c
```diff
@@ -253,16 +300,41 @@ static error_t target_flash_erase_chip(void)
...
         error_t status = ERROR_SUCCESS;
-        const program_target_t *const flash = g_board_info.target_cfg->flash_algo;
-
-        status = flash_func_start(FLASH_FUNC_ERASE);
-        if (status != ERROR_SUCCESS) {
-            return status;
-        }
-        if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
-            return ERROR_ERASE_ALL;
+        program_target_t * flash, * start_flash_algo = last_flash_algo;
```
flit: Please put each variable declaration on a separate line, and init `flash`.

source/daplink/interface/target_flash.c
```diff
@@ -253,16 +300,41 @@ static error_t target_flash_erase_chip(void)
...
+                    }
+                    index++;
+                    break;
+                }
+                index++;
+            }    
+        } while (flash != last_flash_algo); //call erase on the next flash algo
+        
+        //reset flash algo if changed
+        if (flash != start_flash_algo) {
```
flit: You can probably leave this out. The current flash algo will be set correctly on the next erase/program.

source/daplink/interface/target_flash.c
```diff
@@ -253,16 +300,41 @@ static error_t target_flash_erase_chip(void)
...
+            flash = last_flash_algo;
+            status = flash_func_start(FLASH_FUNC_ERASE);
+            if (status != ERROR_SUCCESS) {
+                return status;
+            }
+            if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
+                return ERROR_ERASE_ALL;
+            }
+            
+            while (!(g_board_info.target_cfg->extra_flash[index].start == 0 && g_board_info.target_cfg->extra_flash[index].end == 0)) {
```
flit: Use a `region_info_t` pointer like mentioned above.

source/daplink/interface/target_flash.c
```diff
@@ -253,16 +300,41 @@ static error_t target_flash_erase_chip(void)
...
-            return status;
-        }
-        if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
-            return ERROR_ERASE_ALL;
+        program_target_t * flash, * start_flash_algo = last_flash_algo;
+        uint8_t index = 0;
+        
+        do{ 
+            flash = last_flash_algo;
+            status = flash_func_start(FLASH_FUNC_ERASE);
```
flit: Haha, you ran into the same thing I did in pyOCD! 😄 Do we assume each flash algo only erases that one region, or performs a full chip erase? There are almost certainly differences in how algos behave here. Eventually we need to add a flag that says whether the algo's erase all will erase the whole chip. But this is good for now, and is exactly what I did for pyOCD.

source/daplink/interface/target_flash.c
```diff
@@ -65,9 +67,33 @@ const flash_intf_t *const flash_intf_target = &flash_intf;
...
+            addr >= g_board_info.target_cfg->extra_flash[index].start &&
+            addr <= g_board_info.target_cfg->extra_flash[index].end) {
+                
+            flash_start = g_board_info.target_cfg->extra_flash[index].start;
+            return g_board_info.target_cfg->extra_flash[index].flash_algo;
+        }
+            
+        index++;
+    }
+    flash_start = g_board_info.target_cfg->flash_start;
```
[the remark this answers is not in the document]
brianesquilona: Added check but used the last flash algo, if  in a region where the flash algo is null.

source/daplink/interface/target_flash.c
```diff
@@ -89,20 +115,41 @@ static error_t flash_func_start(flash_func_t func)
...
+        
         if (0 == target_set_state(RESET_PROGRAM)) {
             return ERROR_RESET;
         }
 
-        // Download flash programming algorithm to target and initialise.
-        if (0 == swd_write_memory(flash->algo_start, (uint8_t *)flash->algo_blob, flash->algo_size)) {
-            return ERROR_ALGO_DL;
+        //try to set the initial flash algo
+        error_t status = target_flash_set(g_board_info.target_cfg->flash_start);
```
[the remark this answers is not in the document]
brianesquilona: Hex programming can jump to any address, this just sets the default flash algo, in case there are multiple flash regions using the default flash algo or g_board_info.target_cfg->flash_algo

source/daplink/interface/target_flash.c
```diff
@@ -65,9 +67,40 @@ const flash_intf_t *const flash_intf_target = &flash_intf;
...
+            if (flash_region->flash_algo) {
+                return flash_region->flash_algo;
+            }else{
+                break;
+            }
+        }
+    }
+    
+    //use the current flash algo by default
+    return current_flash_algo;
```
flit: Sorry if I wasn't clear… I meant that if the address doesn't match a flash algo, we should ultimately error out rather than trying to use an algo that may or may not work. So just return NULL from here and check it where called.

source/daplink/interface/target_flash.c
```diff
@@ -253,15 +307,31 @@ static error_t target_flash_erase_chip(void)
...
-            return status;
-        }
-        if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
-            return ERROR_ERASE_ALL;
-        }
+        program_target_t * flash;
+        region_info_t * flash_region = g_board_info.target_cfg->extra_flash;
+        
+        do { 
+            flash = current_flash_algo;
```
flit: I think you need to set `flash` to the main flash algo here, not the current one. Otherwise it's possible to miss some regions.

And you know, I'm really wondering if we should just merge all the flash and ram regions into a single array of each, instead of having a main region and then extra regions array. It would make this code simpler. But don't worry about that for now, maybe later…

flit:
That's all. Just those two additional comments and it's ready.

source/daplink/interface/target_flash.c
```diff
@@ -253,15 +307,38 @@ static error_t target_flash_erase_chip(void)
...
-            return status;
-        }
-        if (0 == swd_flash_syscall_exec(&flash->sys_call_s, flash->erase_chip, 0, 0, 0, 0)) {
-            return ERROR_ERASE_ALL;
-        }
+        program_target_t * flash;
+        flash_region_info_t * flash_region = g_board_info.target_cfg->extra_flash;
+        
+        do {
+            if (current_flash_algo != g_board_info.target_cfg->flash_algo){
```
flit: Since this is in your **do** loop, it will switch to the main flash algo every time you select an extra region algo using the **while** on line 330.

brianesquilona: Oh yeah totally misplaced this! Thanks

PR #616, opened 2019-03-26
Add Musca-A port to DAPLink

avinashmehtadelhi:
Change-Id: I2cfbdd18084344745ae37e07b8e195a716ffc1ee
Signed-off-by: Avinash Mehta <avinash.mehta@arm.com>


source/daplink/interface/main.c
```diff
@@ -363,6 +366,31 @@ void main_task(void * arg)
...
+                    // Loop till PBON is pressed
+                    while (gpio_get_pbon_btn()) {;}
+                    // Power button released when target was already powered off
+                    target_set_state(POWER_ON);
+                    power_on = 1;
+                    reset_pressed = 0;
+                }
+            }
+#endif
+
```
brianesquilona: Is this logic really necessary? Daplink should not care about powering off and on the target in my opinion

source/hic_hal/target_config.h
```diff
@@ -74,6 +78,14 @@ typedef struct target_cfg {
...
     const char *rt_board_id;                                /*!< If assigned, this is a flexible board ID */
     uint16_t rt_family_id;                                     /*!< If assigned, this is a flexible board ID */
+#ifdef TRUSTZONE_M
+    uint32_t flash_start_alias;     /*!< Alias Address of the application entry point */
+    uint32_t flash_end_alias;       /*!< Alias Address where the flash ends */
+    uint32_t ram_start_alias;       /*!< Lowest contigous RAM Alias address the application uses */
+    uint32_t ram_end_alias;         /*!< Highest contigous RAM Alias address the application uses */
+    flash_region_info_t extra_flash_alias[MAX_EXTRA_FLASH_REGION + 1]; /*!< Alias Extra flash regions. */
+    ram_region_info_t extra_ram_alias[MAX_EXTRA_RAM_REGION + 1];       /*!< Alias Extra RAM regions. */
+#endif
```
brianesquilona: I'm not sure why the current api is not enough for the extra ram and flash

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -143,7 +143,15 @@ error_t flash_decoder_get_flash(flash_decoder_type_t type, uint32_t addr, bool a
...
             if (g_board_info.target_cfg) {
+#ifdef TRUSTZONE_M
+                if(addr >= g_board_info.target_cfg->flash_start_alias) {
+                    flash_start_local = g_board_info.target_cfg->flash_start_alias;
+                } else {
+                    flash_start_local = g_board_info.target_cfg->flash_start;
+                }
+#else
                 flash_start_local = g_board_info.target_cfg->flash_start;
+#endif
```
brianesquilona: TODO:

Add patch for multi regions instead of this logic.

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -347,11 +355,19 @@ static bool flash_decoder_is_at_end(uint32_t addr, const uint8_t *data, uint32_t
...
         case FLASH_DECODER_TYPE_TARGET:
             //only if we are sure it is a bin for the target; without check unordered hex files will cause to terminate flashing
             if (flash_type_target_bin && g_board_info.target_cfg) {
+#ifdef TRUSTZONE_M
+                if(addr >= g_board_info.target_cfg->flash_start_alias) {
+                    end_addr = g_board_info.target_cfg->flash_end_alias;
+                } else {
+                    end_addr = g_board_info.target_cfg->flash_end;
+                }
+#else
```
brianesquilona: TODO:

Add patch for multi regions instead of this logic.

source/daplink/interface/main.c
avinashmehtadelhi: There is power on/off sequence which needs to be executed, hence the logic

source/hic_hal/target_config.h
avinashmehtadelhi: v8M introduces secure/non secure alias for the same physical memory/location. extra ram and flash is for different ram and flash locations. musca_B will have two flashes and correcponding 4 s/ns alias

source/daplink/interface/main.c
avinashmehtadelhi: there is power on/off sequence that needs to be executed, hence the logic as PBON button is just another button connected to lpc11u35

source/daplink/drag-n-drop/flash_decoder.c
flit: Please remove this change and rely on existing multiple flash region support.

source/daplink/interface/main.c
flit: We do not allow target-specific code in common files. You'll have to add a clean hook mechanism so you can implement the power button support in your board file.

Although as a user, I'd much prefer the power button was not required. It took a while for me to understand I had to press the button to power up the target when I first got my Musca-A1 board. At the very least, it should automatically power up the target if the CMSIS-DAP debug connection is enabled.

source/daplink/interface/swd_host.c
```diff
@@ -170,7 +170,11 @@ uint8_t swd_read_ap(uint32_t adr, uint32_t *val)
     uint8_t tmp_in, ack;
     uint8_t tmp_out[4];
     uint32_t tmp;
+#ifdef MUSCA_A_BOARD
```
flit: Please add a clean method for configuring the required APSEL, as target-specific logic is not allowed in this common source file. I recommend adding an API to swd_host to set the current APSEL, a field to `target_family_descriptor_t`, and logic to `init_family()` that sets the APSEL based on the selected family.

source/hic_hal/gpio.h
```diff
@@ -42,6 +42,9 @@ void gpio_set_cdc_led(gpio_led_state_t state);
 void gpio_set_msc_led(gpio_led_state_t state);
 uint8_t gpio_get_reset_btn_no_fwrd(void);
 uint8_t gpio_get_reset_btn_fwrd(void);
+#ifdef PBON_BUTTON
```
flit: Remove common code change or replace with a generic API.

source/hic_hal/nxp/lpc11u35/gpio.c
```diff
@@ -152,6 +152,14 @@ uint8_t gpio_get_reset_btn_fwrd()
     return LPC_GPIO->PIN[PIN_RESET_IN_FWRD_PORT] & PIN_RESET_IN_FWRD ? 0 : 1;
 }
 
+#ifdef PBON_BUTTON
```
flit: Move to your board file.

source/hic_hal/target_config.h
```diff
@@ -28,6 +28,10 @@
 #include "flash_blob.h"
 #include "macro.h"
 
+#ifdef IO_CONFIG_OVERRIDE
```
flit: Why is it necessary to include `IO_Config_Override.h` here?

source/hic_hal/target_config.h
flit: Please use the existing extra regions instead of adding more. If needed, add a flags field to `flash_region_info_t` and `ram_region_info_t` structs.

The aliasing is not intrinsic to v8-M, it's simply a recommendation from Arm for how to implement the IDAU given the strict timing requirements.

source/daplink/drag-n-drop/flash_decoder.c
avinashmehtadelhi: Will remove it.

source/daplink/drag-n-drop/flash_decoder.c
avinashmehtadelhi: Will remove this change

brianesquilona:
@avinashmehtadelhi, target apis just been merged. Please rebase and follow the code review by @flit

brianesquilona:
#636 Added both Musca A and B. Can this be closed?

avinashmehtadelhi:
closing this PR as https://github.com/ARMmbed/DAPLink/pull/636 covers both musca A and B port.

PR #620, opened 2019-03-29
Modified target_cfg api

brianesquilona:
1. Array of flash and ram regions are defined
2. Added flash algo, flags, alias_index for memory region
3. Moved target_config.h and target_rest.h to target folder
4. Modified families and bootloader file for new target_config api


source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -143,7 +143,13 @@ error_t flash_decoder_get_flash(flash_decoder_type_t type, uint32_t addr, bool a
             }
         } else if (FLASH_DECODER_TYPE_TARGET == type) {
             if (g_board_info.target_cfg) {
-                flash_start_local = g_board_info.target_cfg->flash_start;
+                region_info_t * region = g_board_info.target_cfg->flash_regions;
+                for (; region->start != 0 || region->end != 0; ++region) {
+                    if (addr >= region->start &&  addr <= region->end) {
```
avinashmehtadelhi: write_bin call this function with addr = 0, if the flash is not starting from addr 0x0 then flashing of bin format will fail, have you tested this? originally flash_start_local used to get populated everytime, but this is not the case anymore, I think the check should be "if (kRegionIsDefault == region->flags) {"

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -337,7 +343,7 @@ error_t flash_decoder_close(void)
 
 static bool flash_decoder_is_at_end(uint32_t addr, const uint8_t *data, uint32_t size)
 {
-    uint32_t end_addr;
+    uint32_t end_addr=0;
```
avinashmehtadelhi: end_addr = 0
spacing

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -351,7 +357,17 @@ static bool flash_decoder_is_at_end(uint32_t addr, const uint8_t *data, uint32_t
...
             if (flash_type_target_bin && g_board_info.target_cfg) {
-                end_addr = g_board_info.target_cfg->flash_end;
+                region_info_t * region = g_board_info.target_cfg->flash_regions;
+                for (; region->start != 0 || region->end != 0; ++region) {
+                    if (addr >= region->start &&  addr<=region->end) {
+                        end_addr = region->end;
+                        break;
+                    }
+                }
+                if(end_addr == 0){ //invalid end_addr
```
avinashmehtadelhi: if (end_addr == 0) { //invalid end_addr
spacing

PR #630, opened 2019-05-14
Flash algo binding for no target builds but basically any build

brianesquilona:
Notes:
1. The flash blob and data structs will be placed at the end of the binary before the crc.
2. Needed a target ram start and end arguments and only support single regions but can easily be extended.
3. Flash blob entry has a default 0x20000000 value but can be specified.
4. Imported flash_algo.py from mbed flash algo script project.

### DIFF
--- a/source/hic_hal/flash_blob.h
+++ b/source/hic_hal/flash_blob.h
@@ -28,13 +28,13 @@
 extern "C" {
 #endif
 
-typedef struct {
+typedef struct __attribute__((__packed__)) {
     uint32_t breakpoint;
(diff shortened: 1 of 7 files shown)


0xc0170:
Can this be split into logical commits ? One commit brings new features, fixes together

brianesquilona:
> Can this be split into logical commits ? One commit brings new features, fixes together

Thanks for the comment @0xc0170 , but in my opinion, the changes are dependent to each other, one change is not significant without the other change.

PR #636, opened 2019-05-22
Add Musca-A & B port to DAPLink

avinashmehtadelhi:


test/info.py
```diff
@@ -1,6 +1,6 @@
 #
 # DAPLink Interface Firmware
-# Copyright (c) 2009-2016, ARM Limited, All Rights Reserved
+# Copyright (c) 2009-2019, ARM Limited, All Rights Reserved
```
0xc0170: Please revert changing permissions to this file (100644 → 100755)

0xc0170: Few files in this PR

source/family/arm/musca_a/target_reset.c
```diff
@@ -0,0 +1,168 @@
...
+    buf[6] = 0x00;
+    buf[7] = 0x10;
+    buf[8] = 0xFE;
+    buf[9] = 0xE7;
+    buf[10] = 0xFE;
+    buf[11] = 0xE7;
+
+    swd_write_memory(0x10000000, (uint8_t *)buf, 12);
+
+//    swd_write_word(0x10000000, 0x30008000);
```
0xc0170: why is this dead code left here?

source/daplink/interface/main.c
```diff
@@ -363,6 +366,29 @@ void main_task(void * arg)
                 reset_pressed = 0;
             }
 
+#ifdef PBON_BUTTON
+            // handle PBON pressed
```
0xc0170: I've seen some comments about `PBON_BUTTON`  - why is this needed here and how this can be used generically ?

test/info.py
avinashmehtadelhi: changed

source/family/arm/musca_a/target_reset.c
avinashmehtadelhi: removed now.

source/daplink/interface/main.c
avinashmehtadelhi: targets with Power button (like musca) can #define PBON_BUTTON in there override file and power button handling will be compiled for those boards.

0xc0170:
@brianesquilona @flit looks fine to me, not certain about the button addition - can you please review?

0xc0170:
Let's progress here.

I still dont like the change for PBON_BUTTON - will get this resolved on master somehow.

source/daplink/interface/main.c
0xc0170: I read the comments in previous PR -is this really needed ? If I remove this ifdef from here - what is a result? Why there is this button check, it's not common funcitonality.

Even with ifdef, this is polluting common code. I would rather have a hook - functionality there to hook into than ifdef. I merged this to progress , but still see this as unresolved

PR #648, opened 2019-09-04
Add support for EP_AGORA

millerc-ep:


millerc-ep:
@loverdeg-ep @trowbridgec

0xc0170:
How was this tested ? It would be helpful to provide logs as part of this PR

40Grit:
@0xc0170 This was done blind since we do not have an ac5 license.

Was hoping a build could be done via CI so we can use artifacts to test on our end.

test/info.py
```diff
@@ -56,6 +56,7 @@
     ("k20dx_mimxrt1020_evk_qspi_if",                True,       0x8000,     "bin"       ),
     ("k20dx_mimxrt1050_evk_hyper_if",               True,       0x8000,     "bin"       ),
     ("k20dx_mimxrt1050_evk_qspi_if",                True,       0x8000,     "bin"       ),
+    ('k20dx_ep_agora_if',                           False,      0x0000,     "bin"       ),
```
trowbridgec: Do we need to match this to the other `k20dx` entries?

For example:
```
('k20dx_ep_agora_if',                           True,      0x8000,     "bin"       ),
```

loverdeg-ep:
@maclobdell

test/info.py
0xc0170: I would say yes

0xc0170:
> Was hoping a build could be done via CI so we can use artifacts to test on our end.

I can do. I'll start CI and share what I can with you

0xc0170:
2 errors in the build:

```
13:10:07 Compile [  1.3%]: ep_agora.c
13:10:07 [Error] ep_agora.c@28,21:  #154: expression must have struct or union type
13:10:07 [Error] ep_agora.c@29,16:  #57: this operator is not allowed in a constant expression
13:10:08 [ERROR] ".\source\board\ep_agora.c", line 28: Error:  #154: expression must have struct or union type
13:10:08       .infoVersion = 0x0002
13:10:08                      ^
13:10:08 ".\source\board\ep_agora.c", line 29: Error:  #57: this operator is not allowed in a constant expression
13:10:08       .board_id = "2600",
13:10:08                 ^
13:10:08 .\source\board\ep_agora.c: 0 warnings, 2 errors
```

test/info.py
```diff
@@ -56,7 +56,7 @@
     ("k20dx_mimxrt1020_evk_qspi_if",                True,       0x8000,     "bin"       ),
     ("k20dx_mimxrt1050_evk_hyper_if",               True,       0x8000,     "bin"       ),
     ("k20dx_mimxrt1050_evk_qspi_if",                True,       0x8000,     "bin"       ),
-    ('k20dx_ep_agora_if',                           False,      0x0000,     "bin"       ),
+    ('k20dx_ep_agora_if',                           True,      	0x0000,     "bin"       ),
```
loverdeg-ep: Probably will want to fix the whitespace. Looks like a TAB too.

40Grit: @0xc0170 looking at this on my phone it doesn't appear like it did on my laptop

@millerc-ep is this ready for another build?

trowbridgec: @millerc-ep Spacing still seems a little off to me, might want to review just so it looks clean.

0xc0170:
I manually restarted the tests. Looks like CI does not report back status, have to check why not.

0xc0170:
The build completed, the EP_AGORA is not in the build artifacts , will investigate

0xc0170:
[k20dx_ep_agora_if.zip](https://github.com/ARMmbed/DAPLink/files/3591507/k20dx_ep_agora_if.zip)


Attached agora artifacts, please verify

trowbridgec:
@0xc0170 Thank you!

The build looks good on our end...I was able to utilize drag and drop (MSD) functionality as well as debugging/flashing via pyocd with the `k20dx_ep_agora_if_crc_legacy_0x5000.bin` artifact on top of bootloader `0244_k20dx_bl_0x5000.bin` from the [here](https://os.mbed.com/blog/entry/DAPLink-bootloader-updates-with-external/).

0xc0170:
PR ready for integration?

loverdeg-ep:
@0xc0170 I think it is ready. I would like to understand what the difference is between the "legacy" and not legacy build though.

0xc0170:
I expect test to fail, Mac machine has a problem with boards these days. I'll review if all the rest is OK

0xc0170:
All tests now green 💯

PR #661, opened 2019-10-21
rtos: fix systick inclusion

### LINKED ISSUE #654: #include does not have the correct case
https://github.com/ARMmbed/DAPLink/blob/master/source/rtos_none/SysTick_Handler.c#L22
https://github.com/ARMmbed/DAPLink/blob/master/source/rtos_none/cmsis_os2_port.c
Fails to compile on case-sensitive systems because Systick_Handler.h != SysTick_Handler.h.

0xc0170:
Fixes #654

### DIFF
--- a/source/rtos_none/SysTick_Handler.c
+++ b/source/rtos_none/SysTick_Handler.c
@@ -19,7 +19,7 @@
  * limitations under the License.
  */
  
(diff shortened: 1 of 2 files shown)


0xc0170:
All green, going in

PR #675, opened 2019-11-21
MIMXRT1050: Update Flash Algo

mmahadevan108:
The old flash blob will add FCB and IVT into image. This version will not add FCB and IVT, only program the raw image.

Signed-off-by: Mahesh Mahadevan <mahesh.mahadevan@nxp.com>


mmahadevan108:
cc @0xc0170

flit:
Thanks!

Issue #197, opened 2016-11-30
Support GCC


Timmmm:
It would be nice if you didn't have to spend thousands on a Keil license to compile this!

kylemanna:
Additional gcc discussion at #110

c1728p9:
Before GCC support can be added RTX needs to be updated. The version of RTX used by DAPLink doesn't have support for GCC, but newer versions do.

flit:
Does the USB stack already support gcc?

c1728p9:
I don't think the USB stack has strong ties to any toolchain, so it should be fairly straight forward to update to use GCC.

OwenBrotherwood:
Can we track who has the knowledge and who has the time to obtain a basic "this is as far as I/we got" to facilitate PR's for doc and possible forks to show their results be they ever so humble with possible PR's to code with possible problems not solved at the moment.

OwenBrotherwood:
@c1728p9  named RTX.
This is probably http://www2.keil.com/mdk5/cmsis/rtx/

> Royalty-Free - No on-going costs.
> There are no run-time royalty payments or other hidden charges. Ship your RTX based products without further fees or recurring costs

Given no cost, and if the RTX refered to is the one I found, the github seems to be a start for compiling?
https://github.com/ARM-software/CMSIS_5

c1728p9:
Hi @OwenBrotherwood, yep that is the right RTX. Since it's open source anyone who is interested could start the porting process.

OwenBrotherwood:
In order to create a form for "if someone suddenly does not have the time", the problem needs to be broken up into blocks that are describable as start/finish such that one can build up to the final result
Anyone have a breakdown on the blocks required?

OwenBrotherwood:
It would seem that the correct branch is this for micro:bit
https://github.com/mbedmicro/DAPLink/tree/freescale

jaustin:
Actually @OwenBrotherwood this is the right place for micro:bit https://github.com/mbedmicro/DAPLink/tree/microbit-0234 (that's the schools drop) and https://github.com/mbedmicro/DAPLink/tree/v0241 (the base that Farnell used for the newer firmware)

OwenBrotherwood:
So many micro:bits out there and so many people wanting to do cool things with them.
We need open-source compilation.

OwenBrotherwood:
Dunno if this merge of code gives anyone any ideas.
[...]

PR #678, opened 2019-11-23
Add experimental GCC ARM toolchain

xiongyihui:
To support GCC ARM toolchain:

1.  convert some attributes

    + [`__weak` to `__attribute__((weak))`](https://github.com/xiongyihui/DAPLink/blob/gccarm/records/tools/make_gcc_arm.yaml#L6)
    + `__forceinline` to `inline __attribute__((always_inline))`
    + `__packed` to `__attribute__((__packed__))`, also need to change its location, for example

      convert `__packed struct { char a; int b; } c;` to `struct __attribute__((__packed__)) { char a; int b; } c;`.
      it is used by USB stack

2.  wide character array defined by `L"a wide string"`

    By default, gcc uses 4 bytes to store a wide char, while armcc uses 2 bytes. In order to be compatible with armcc, add `-fshort-wchar` to gcc compiler flags.

3. convert asm code

Right now, it work partially. To test it.
```
rm source/daplink/interface/SVC_Table.s
progen generate -f projects.yaml -p lpc11u35_archble_if_bm -t make_gcc_arm
cd projectfiles/make_gcc_arm/lpc11u35_archble_if_bm
make
```

### DIFF
--- a/projects.yaml
+++ b/projects.yaml
@@ -53,6 +53,9 @@ module:
     hic_lpc11u35: &module_hic_lpc11u35
         - records/rtos/rtos-cm0.yaml
         - records/hic_hal/lpc11u35.yaml
+    hic_lpc11u35_bm: &module_hic_lpc11u35_bm
+        - records/rtos/rtos-none.yaml
+        - records/hic_hal/lpc11u35.yaml
     hic_lpc4322: &module_hic_lpc4322
         - records/rtos/rtos-cm3.yaml
         - records/hic_hal/lpc4322.yaml
@@ -321,6 +324,10 @@ projects:
     lpc11u35_archble_if:
         - *module_if
         - *module_hic_lpc11u35
+        - records/board/archble.yaml
+    lpc11u35_archble_if_bm:
+        - *module_if
+        - *module_hic_lpc11u35_bm
         - records/board/archble.yaml
     lpc11u35_tiny_if:
         - *module_if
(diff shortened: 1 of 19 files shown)


xiongyihui:
@0xc0170 Do you have any idea that the gccarm firmware shows an empty disk? Which part of code is related?


image

brianesquilona:
> @0xc0170 Do you have any idea that the gccarm firmware shows an empty disk? Which part of code is related?
> 
> image

@xiongyihui you might want to check all msc related codes in source/usb.

xiongyihui:
I figured out the issue. It is that `cfgrom` and `cfgram` are not configured at the linker script.

xiongyihui:
It seem armcc uses much less RAM and Flash than gccarm. LPC11U35 doesn't have enough to run the interface firmware compiled with gccarm, even if without RTX RTOS.
A workaround is to make [`target_device` to a const variable](https://github.com/ARMmbed/DAPLink/blob/master/source/target/target_config.h#L77)

flit:
See the latest on #197. Unfortunate timing, but I'm almost done with the official port to gcc.

xiongyihui:
I see. I will test that.

projects.yaml
```diff
@@ -53,6 +53,9 @@ module:
     hic_lpc11u35: &module_hic_lpc11u35
         - records/rtos/rtos-cm0.yaml
         - records/hic_hal/lpc11u35.yaml
+    hic_lpc11u35_bm: &module_hic_lpc11u35_bm
```
0xc0170: As these might not be related to GCC port, please send t hem separately  or after GCC support lands on master.

0xc0170:
@xiongyihui Thanks for the effort ! Please review and test https://github.com/flit/DAPLink/tree/feature/gcc

PR #680, opened 2019-11-26
Fix numerous little issues

flit:
- The CMSIS register definitions header for STM32F103xB has duplicate definitions of Cortex-M PPB peripherals such as NVIC, SysTick, and SCB.
- Using DAPLink's `util_assert()` for Kinetis drivers.
- Removed unsupported family and board sources.
- Removed some `target_reset.c` sources that did nothing.
- Removed `board_init()` function that was mostly unused.
- Renamed all cases of static functions whose names conflicted with global function names in board and family sources.
- Fixed `sectors_info` array type for STM32F407VE.
- Fixed missing include of `swd_host.h` in wizwiki board sources.
- Corrected functions that were declared extern in `swd_host.h` but defined as static in `swd_host_ca.c`.


0xc0170:
Thanks @flit . I'll review later today

py3 fixes included 💯

0xc0170:
Nice clean up, merging now

PR #688, opened 2020-01-06
More misc cleanup

flit:
- Cleanup of `main.h` headers.
- Merged `macro.h` into `util.h`
- Moved `ARRAY_SIZE()` macro to `util.h`, replacing `ELEMENTS_IN_ARRAY()`.
- Using a single definition of `MAX()` from `util.h`.
- Changed the `disc_size` constant in `vfs_user.c` to a macro.
- Replaced the weak, null family descriptor structs with weak references to those structs and a different terminator value for `g_families` (so entries in that array can be null without terminating the list).
- Changed most includes of C Standard Library headers to be system includes (angle brackets instead of quotes). Didn't touch vendor-supplied SDK sources for the most part.
- Trailing whitespace cleanup.

PR #689, opened 2020-01-10
Cleanup board info version

flit:
The `board_info_t::infoVersion` member was intended to be used as the version number for the `board_info_t` struct itself, not the version of the board. This is important to ensure that board info added via post-processing a firmware binary has a known format. This PR clarifies and standardizes the usage of this version member.

Defined a `kBoardInfoVersion` constant (currently equal to 1), and set it as the current version in all defined `board_info_t` structures. Also renamed `infoVersion` to `info_version`. The interface main task asserts that `g_board_info.info_version` is valid.

(Extracted from #688.)

### DIFF
--- a/source/board/6LoWPAN_BorderRouterETHERNET.c
+++ b/source/board/6LoWPAN_BorderRouterETHERNET.c
@@ -23,7 +23,7 @@
 #include "target_family.h"
 
 const board_info_t g_board_info = {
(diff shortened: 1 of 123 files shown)


flit:
Mac and Windows CI still have the unavailable board issue, but Linux passed so I'm merging.

PR #691, opened 2020-01-12
Refactor magic file handling

flit:
This PR encompasses two changes:

1. Use a table to define the set of standard magic files.
2. Add weak hook functions for user file change notifications and to override or extend magic files.


0xc0170:
@flit Based on the failures, looks like `sam3u2c_mkit_dk_dongle_nrf5x_if` is not available (the rest looks fine), does this board need to be restarted?

flit:
Yes, the sam3u2c_mkit_dk_dongle_nrf5x board probably needs to be replugged. I'll check if Brian can do that today.

PR #692, opened 2020-01-13
Change weak target family descriptors to references

flit:
This change replaces the default weak family descriptor structs (each taking up the full family descriptor size) in `target_family.c` with weak references. The purpose is to save `.rodata` space.

### DIFF
--- a/source/target/target_family.c
+++ b/source/target/target_family.c
@@ -43,32 +43,32 @@ const target_family_descriptor_t g_sw_sysresetreq_family = {
     .soft_reset_type = SYSRESETREQ,
 };
 
-//Weakly define family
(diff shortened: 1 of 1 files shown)


0xc0170:
Can you check test failures to proceed here?

mathias-arm:
I think this PR has gotten stale. A lot has been merged to develop, but changing
```
__attribute__((weak)) const target_family_descriptor_t g_xxxxx  = {0};
```
to
```
extern __WEAK const target_family_descriptor_t g_xxxx;
```
caused some issue with `armcc` so it was backed out (see [fb4b2292981671c6441d08eec8e51c3a685f4390]).

PR #696, opened 2020-01-22
Make the JTAG2SWD function non-static

MarianSavchuk:
I would like to call the JTAG2SWD() function from the target_before_init_debug() function


flit:
Would you mind rebasing and squashing to clean up the merge commits? (Force push to avoid having to create a new PR.) Thanks!

PR #697, opened 2020-01-22
place CMSIS-DAP v2 to the interface #0

MarianSavchuk:
This enables DAPLink detection with all versions of uVision and IAR EWB that support CMSIS-DAPv2

### DIFF
--- a/source/usb/usb_def.h
+++ b/source/usb/usb_def.h
@@ -304,6 +304,15 @@ typedef __packed struct _USB_BINARY_OBJECT_STORE_DESCRIPTOR {
     U8  bNumDeviceCaps;
 } USB_BINARY_OBJECT_STORE_DESCRIPTOR;
 
+/* Call Management Functional Descriptor */
+typedef __packed struct _CALL_MANAGEMENT_FUNCTIONAL_DESCRIPTOR {
+    U8  bLength;
+    U8  bDescriptorType;
+    U8  bDescriptorSubtype;
+    U8  bmCapabilities;
+    U8  bDataInterface;
+} CALL_MANAGEMENT_FUNCTIONAL_DESCRIPTOR;
+
 /* Union Functional Descriptor */ 
 typedef __packed struct _UNION_FUNCTIONAL_DESCRIPTOR {
     U8  bLength;
--- a/source/usb/usb_lib.c
+++ b/source/usb/usb_lib.c
@@ -2547,7 +2547,9 @@ static U16 acm_cdc_desc_fill(U8 * config_desc, U8 * config_desc_hs, U8 if_num) {
 #endif    
     
     ((USB_INTERFACE_DESCRIPTOR *)pD)->bInterfaceNumber = if_num;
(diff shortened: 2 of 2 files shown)


flit:
This configuration was our initial intent. Unfortunately, it breaks the "Mbed Serial driver", required for Windows 7 (and actually a composite driver + serial driver), with a BSOD.

We looked at updating the driver, but it was written long, long ago by a contractor and the source has been lost. And (sadly, imo) there are still a rather large number of users that use Windows 7, especially in Asia. The choice was between a relatively minor bug in µVision (lost functionality) that can be fixed versus a non-fixable whole-system crash. (Also, on Windows 7 if MSC is not interface #0 then it will not be detected by the OS unless the Mbed Serial driver is installed.)

Note that Keil µVision has been updated to fix the detection issue.
Does IAR EWB also have the same or similar issue? We should work with them to get the bug fixed.

I'll not close this issue yet; I'd love to have a better fix and I'd like feedback from the other owners. (Also to double-check my understanding of the issue.) One potential solution may simply be a compile-time configuration option (macro) that reorders the interfaces, so you can place CMSIS-DAPv2 first if you don't need Win7 support.

cc @brianesquilona @0xc0170

MarianSavchuk:
@flit, 
Probably I found a combination of drivers when DAPLink works without the problems you described.
USB Composite Device - HWID:USB\VID_0D28&PID_0204&REV_0100, Native Windows driver (v.6.1.7601.24138)
CMSIS-DAP v2 - HWID:USB\VID_0D28&PID_0204&REV_0100&MI_00, inf file from https://arm-software.github.io/CMSIS_5/DAP/html/group__DAP__ConfigUSB__gr.html#WinUSB
USB Mass Storage Device - HWID:USB\VID_0D28&PID_0204&REV_0100&MI_01, Native Windows driver (v.6.1.7601.19144)
mbed Serial Port - HWID:USB\VID_0D28&PID_0204&REV_0100&MI_02, Mbed Serial driver (v.2.0.2.0)

brianesquilona:
@flit is correct recent uvision should have patched this bug (bulk end detection problem if it's not the first interface). We can remove some of the interfaces via yaml but the order currently is by the code.

0xc0170:
> We looked at updating the driver, but it was written long, long ago by a contractor and the source has been lost. And (sadly, imo) there are still a rather large number of users that use Windows 7, especially in Asia. The choice was between a relatively minor bug in µVision (lost functionality) that can be fixed versus a non-fixable whole-system crash. (Also, on Windows 7 if MSC is not interface #0 then it will not be detected by the OS unless the Mbed Serial driver is installed.)

I've got the sources thanks to ChrisS. I was able to compile them and quick test on one machine. Although not yet familiar with the sources much, if I know what I should look for, I can have a check in the sources and might be able to recompile the driver if that would help? Would it be the best approach here to fix the problem?

0xc0170:
if uvision was fixed, IAR potentially could be also if not yet. What else is there to fix here?

MarianSavchuk:
@0xc0170, could you please consider this approach, which is mentioned in the first comment by @flit.

MarianSavchuk:
I am closing this request because it is no longer relevant.

PR #698, opened 2020-01-22
expand DAP Vendor Command IDs to range 0xA0U-0xFEU

MarianSavchuk:
The range 0x80U-0x9FU of DAP Vendor Command IDs is reserved for generic DAPLink purposes.
For specific Vendor features, the range of DAP command identifiers has been expanded to the range 0xA0U-0xFEU

### DIFF
--- a/source/daplink/cmsis-dap/DAP.h
+++ b/source/daplink/cmsis-dap/DAP.h
@@ -1,5 +1,7 @@
 /*
  * Copyright (c) 2013-2019 ARM Limited. All rights reserved.
+ * Copyright 2019, Cypress Semiconductor Corporation 
+ * or a subsidiary of Cypress Semiconductor Corporation.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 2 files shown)


source/daplink/cmsis-dap/DAP.h
```diff
@@ -101,6 +103,104 @@
...
+#define ID_DAP_Vendor117                0xF5U
+#define ID_DAP_Vendor118                0xF6U
+#define ID_DAP_Vendor119                0xF7U
+#define ID_DAP_Vendor120                0xF8U
+#define ID_DAP_Vendor121                0xF9U
+#define ID_DAP_Vendor122                0xFAU
+#define ID_DAP_Vendor123                0xFBU
+#define ID_DAP_Vendor124                0xFCU
+#define ID_DAP_Vendor125                0xFDU
+#define ID_DAP_Vendor126                0xFEU
```
flit: I'm not sure we need to predefine all these command IDs. It might be better to define only `ID_DAP_VendorExFirst` and `ID_DAP_VendorExLast` so the range is documented in the code. (Note the "Ex" in the name to match `DAP_ProcessVendorCommandEx()`.) Then any vendor commands should have their own definitions in a HIC-specific header file, something like `#define ID_DAP_CypressFoo (ID_DAP_VendorExFirst + 1)

flit:
Please rebase, squash, and force push to clean up the merge commits. Thanks!

source/daplink/cmsis-dap/DAP.h
MarianSavchuk: I've made changes according to your comments.

PR #700, opened 2020-01-27
flashalgo flags are introduced:

MarianSavchuk:
kAlgoVerifyReturnsAddress - Verify function returns address if this flag is set
kAlgoSingleInitType - The Init/UnInit functions contain just return, so can be skipped
kAlgoSkipChipErase - Skip region when erase.act action triggers. ChipErase on Some SPI Flash Memory take much more time that is allowed per drag-n-drop programming.

### DIFF
--- a/source/daplink/interface/swd_host.c
+++ b/source/daplink/interface/swd_host.c
@@ -4,6 +4,8 @@
  *
  * DAPLink Interface Firmware
  * Copyright (c) 2009-2019, ARM Limited, All Rights Reserved
+ * Copyright 2019, Cypress Semiconductor Corporation 
+ * or a subsidiary of Cypress Semiconductor Corporation.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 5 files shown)


0xc0170:
Can you please describe this change at least in few words?  It helps reviewers to get an idea what this introduces.

MarianSavchuk:
Description of the purpose of the Flashalgo flags

_kAlgoVerifyReturnsAddress_ 

Flash programming algorithms based on https://arm-software.github.io/CMSIS_5/Pack/html/flashAlgorithm.html are different from algorithms based on https://github.com/mbedmicro/FlashAlgo.
In CMSIS-Pack based, the Verify function returns the status which is the start address + size of the verification on success, any other number - on failure, and represents the failing address.
In flash algorithm based on the FlashAlgo project, the Verify function returns 1 on success and 0 on failure.
If kAlgoVerifyReturnsAddress flag is set it mean that Verify function returns the start address + size of the verification on success

_kAlgoSingleInitType_

Some flash algorithms have Init/Uninit functions that ignore arguments, so there is no need to call them from flash_func_start every time when current flash function changes among FLASH_FUNC_ERASE, FLASH_FUNC_PROGRAM and FLASH_FUNC_VERIFY.
If kAlgoVerifyReturnsAddress flag is set it mean that the Init/Uninit functions ignore their arguments and may be safely skipped.

_kAlgoSkipChipErase_

When processing the file "erase.act", iteration of all defined flash regions is performed and for each erase_chip will be executed.
And erase_chip will be called even if DAPLink is created with page erase enabled.
For external SPI flash memory (e.g. S25FL512S) erase_chip usually last around 100 seconds.
On the other hand, the operating system imposes restrictions on the execution time of an I/O operation on the MSC. 
For example, Windows will wait about 20 seconds and then perform reset the USB bus for the USB device.
So such memory cannot be erased in this way and should be skipped.
If kAlgoSkipChipErase flag is set it mean that the corresponding flash region cannot be erased by processing "erase.act" file.

PR #743, opened 2020-11-24
always put the main.o code and the startup data in RAM1

groleo:
When DAPLINK_RAM_APP2_START is defined, the startup code and data
may be placed in RAM2, which is not guaranteed to work.

Explicitly place main.o and startup_*.o in RAM1.

Signed-off-by: Adrian Negreanu <adrian.negreanu@nxp.com>

### DIFF
--- a/source/daplink/daplink.sct
+++ b/source/daplink/daplink.sct
@@ -33,6 +33,8 @@ LR_IROM1 DAPLINK_ROM_APP_START DAPLINK_ROM_APP_SIZE {
   RW_IRAM1 DAPLINK_RAM_APP_START DAPLINK_RAM_APP_SIZE {  ; RW data
    .ANY (ram_func)
    .ANY (+RW +ZI)
+   main.o (+RW +ZI)
+   startup_*.o (+RW +ZI)
    .ANY (RAM1)
   }
 


groleo:
cc @flit @maclobdell @marcuschangarm @0xc0170

0xc0170 (approved):

source/daplink/daplink.sct
```diff
@@ -33,6 +33,8 @@ LR_IROM1 DAPLINK_ROM_APP_START DAPLINK_ROM_APP_SIZE {
   RW_IRAM1 DAPLINK_RAM_APP_START DAPLINK_RAM_APP_SIZE {  ; RW data
    .ANY (ram_func)
    .ANY (+RW +ZI)
+   main.o (+RW)
+   startup_*.o (+ZI)
```
flit, line 37: Any reason you didn't use `(+RW +ZI)` for both `main.o` and `startup_*.o`? Presumably the excluded type isn't required by the code right now, but it might be in the future and it would be a painful thing to debug.

groleo: I wanted to pin down only the minimum needed.
I'll give it a go with both RW and ZI.
If it's fine, I'll update this PR.

groleo:
v2: use (+RW +ZI) for both main.o and startup_*.o

flit (approved):

PR #746, opened 2020-12-05
Update pyocd and progen versions

flit:
Update pyocd to 0.28.* and progen to 0.10.*.

- Updated pyocd API usage in test suite.
- Removed flash page programming test from `hid_test.py` since it doesn't test anything more than does initially programming the test firmware.


0xc0170:
Now pending!

0xc0170 (approved):

PR #754, opened 2020-12-21
KL27Z base port

microbit-carlos:
Submitting this on behalf of @gerargz, who is on well deserved holiday.

This PR  is for the base DAPLink port for the microbit v2 board with the KL27Z HIC. This branch doesn't have any custom feature code.

PR #760, opened 2020-12-31
KL27Z base port merge

mathias-arm:
Merge of #754 (keeping commit IDs)


flit (approved):

mathias-arm:
@0xc0170 @flit @brianesquilona: test was successful on `linux` but failed on `win7` because it could not test `kl26z_microbit_if`.

Issue #719, opened 2020-06-11
details.txt in the latest master doesn't expand all the URL


microbit-carlos:
Building from the latest commit in master d15cc3d2b39c550842f658593e4b7980f6e06148 causes issues in the details.txt file and it doesn't expand all the URL parameters.

Building the micro:bit interface and loading it into a micro:bit v1.5 (which has bootloader 0246) produces the following details.txt:

```
# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 9901360149624e45005b400f000000250000000097969901
HIC ID: 97969901
Auto Reset: 1
Automation allowed: 0
Overflow detection: 0
Page erasing: 0
Daplink Mode: Interface
Interface Version: 0254
Bootloader Version: 0246
Git SHA: d15cc3d2b39c550842f658593e4b7980f6e06148
Local Mods: 0
USB Interfaces: MSD, CDC, HID, WebUSB
Bootloader CRC: 0xa28dc05a
Interface CRC: 0x376dd06a
Remount count: 0
URL: https://microbit.org/device/?id=9901&v=@V
```

You can see the URL is `https://microbit.org/device/?id=9901&v=@V` instead of `URL: https://microbit.org/device/?id=9901&v=0254`

flit:
Thanks for the bug report.

mathias-arm:
We are hitting the 512 bytes limit. It would be complicated to support longer files.

We could discuss removing a few lines (either unilaterally or conditionally, e.g. when the value is 0), or move some of the content to a distinct file.

mathias-arm:
A few points that have been discussed:
- The URL field is in the HTML file and it seems to serve no real purpose here.
- The "Git SHA" does not need to be a full hash.
- The URL at the top is not working so the situation should be fixed or it should be removed.

I was thinking we could remove the URL and compress "Git SHA", "Local mods", and the newer "Compiler" into a single `Build: d15cc3d (modified, armcc)`. That could leave some space to possibly add the name of the firmware, e.g. `kl26z_microbit_if` or part(s) of it.

mathias-arm:
While modifying we are modifying are there other things that should be changed? Some HICs now have support for SWO and JTAG support, should we expose that information?

microbit-carlos:
There is current the `BOARD_EXTRA_BUFFER` method of increasing the file buffer. Is this still required? Do we need to use it in the micro:bit v1 configuration?

mathias-arm:
[...]

Issue #714, opened 2020-06-02
LED activity during WebUSB DAPJs flashing doesn't blink in 0254


microbit-carlos:
Raising this issue here as the same version of DAPJs works fine with DAPLink 0253, but doesn't with 0254, but maybe this issue has to be transferred.  

Basically, flashing with DAPJs via WebUSB is meant to blink the DAPLink LED, and this stopped working with DAPLink 0254.

Can be replicated using the DAPJs demo and a micro:bit board:
- https://armmbed.github.io/dapjs/examples/daplink-flash/web.html
- Hex to show heart in the micro:bit LED matrix: [makecodeheart.hex.zip](https://github.com/ARMmbed/DAPLink/files/4717966/makecodeheart.hex.zip)
- DAPLink interface 0253: [0253_kl26z_microbit_0x8000.hex.zip](https://github.com/ARMmbed/DAPLink/files/4717979/0253_kl26z_microbit_0x8000.hex.zip)
- DAPLink interface 0254: [0254_kl26z_microbit_0x8000.hex.zip](https://github.com/ARMmbed/DAPLink/files/4717971/0254_kl26z_microbit_0x8000.hex.zip)

microbit-carlos:
@thegecko tagging you to this ticket for visibility, as I am still not quite sure if this is a fix in DAPJs or DAPLink.

microbit-carlos:
Quick note to add that serial activity via WebUSB doesn't blink the LED either in 0254.

thegecko:
I agree this sounds like a DAPLink issue,  with your repro steps it should really help tie down the change which introduced this regression

thegecko:
Also worth mentioning we use the latest DAPjs with older DAPLink firmware and don't see this issue.

microbit-carlos:
@flit a partial fix for this issue when WebUSB flashing was included as part of PR #763 in this commit: https://github.com/ARMmbed/DAPLink/pull/763/commits/e4e9b2d8b8fcc711fcfe9c4858701548643c825f

However, the LED still doesn't blink during "partial flashing". In the MakeCode and Python editor we use DAPJs with WebUSB CMSIS-DAP to manually erase/flash specific pages to update the user code only ([JS code here](https://github.com/bbcmicrobit/PythonEditor/blob/v2.1.2/js/partial-flashing.js)). With older versions of DAPLink the LED would flash during partial flashing as well, but it currently doesn't.

flit:
[...]

PR #763, opened 2021-01-06
Microbit v2 specific features

gerargz:
Draft PR with microbit v2 specific features. This contains the commit ID for the DAPLink build used by the factory release of the microbit v2 boards. 

This PR is still a draft as it contains modifications to the DAPLink core that may not be upstream friendly.


source/board/microbitv2/microbitv2.c
```diff
@@ -18,23 +19,591 @@
...
+    if (!memcmp(filename, "ERASE   ACT", sizeof(vfs_filename_t))) {
+        // Erase last 128KB flash block
+        FLASH_Erase(&g_flash, FLASH_CONFIG_ADDRESS, g_flash.PFlashTotalSize / g_flash.PFlashBlockCount, kFLASH_apiEraseKey);
+    }
+    
+    return false;
 }
 
-// USB HID override function return 1 if the activity is trivial or response is null 
+void vfs_user_build_filesystem_hook() {
```
microbit-carlos, line 547: When DAPLink encounters an irrecoverable error we erase the target flash and flash a small program on the nRF to scroll the error code on the display.

The error codes come from an enum in `source/daplink/error.h`: https://github.com/microbit-foundation/DAPLink-microbit/blob/d7f24a673dc417fd169bd1f2ae196eb5a796303b/source/daplink/error.h#L30-L85

The enum is grouped in categories, so if future DAPLink updates add extra errors in one of the categories before the end of the enum, then all subsequent error values will shift.

It's quite important for us that the values shown to the user don't change, to be able to maintain consistency and so that online documentation of the error codes is always correct.
I guess the only way to "fix in place" the error values would be to only add new errors at the end of the enum. Is this something you would consider? Otherwise we might need to convert these values here, but how that is done would depend on the changes in `error_t` and it could be easy to accidentally update one without the other.

source/daplink/cmsis-dap/DAP_vendor.c
```diff
@@ -101,6 +101,9 @@ uint32_t DAP_ProcessVendorCommand(const uint8_t *request, uint8_t *response) {
         // uart read
         int32_t read_len = 62;
         read_len = uart_read_data(response + 1, read_len);
+        if (read_len) {
+            main_blink_cdc_led(MAIN_LED_FLASH);
+        }
```
microbit-carlos, line 106: The changes in this file were introduced as a fix for https://github.com/ARMmbed/DAPLink/issues/714

source/daplink/cmsis-dap/DAP.c
```diff
@@ -1791,5 +1789,22 @@ void DAP_Setup(void) {
...
+    if (delay > IO_PORT_WRITE_CYCLES) {
+      delay -= IO_PORT_WRITE_CYCLES;
+      delay  = (delay + (DELAY_SLOW_CYCLES - 1U)) / DELAY_SLOW_CYCLES;
+    } else {
+      delay  = 1U;
+    }
+
+  DAP_Data.clock_delay = delay;
+  }
+
```
microbit-carlos, line 1808: A bit more info on this change from @gerargz:


> #### HIC Specific
> Changed the way the KL27 accesses the GPIO for SWCLK and SWDIO. Using the IOPORT interface to access the GPIO (FGPIO address) makes this a single cycle access instead of the three cycles it takes when accessing the GPIO through the peripheral bridge. This change now matches with IO_PORT_WRITE_CYCLES definition.
> 
> Changed the default SWCLK speed for the KL27 HIC from 5MHz to 8MHz (nrf max).
> 
> #### DAPLink core
> Changed the default way that DAPLink calculates the initial clock delay to achieve the desired SWCLK speed.
> 
> #### Board specific
> Added macro DELAY_FAST_CYCLES=2U to the microbit yaml file to allow 8MHz calculation.
> 
> ### Background
> DAPLink inserts delays to approximate the SWCLK speed to the one configured. There are two methods for inserting these delays, the “fast” delay and “slow” delay. The fast delay uses a fixed number of precompiled “nop” instructions (up to 3) and the slow delay uses a loop to have a wider range of clock delays but this has a 3 cycle granularity per loop.
> 
> In DAP_Setup() the clock_delay is calculated without considering that the “slow” delay takes 3 cycles for each iteration, so the resulting clock speed will have extra delay. In contrast, when using a DAP command (DAP_SWJ_Clock()) to change the SWCLK speed, the clock_delay does consider the extra cycles for the “slow” delay and is calculated correctly. In this PR, the method of calculating the clock delay from DAP_SWJ_Clock() was added to DAP_Setup().
> 
> To generate an 8MHz SWCLK for the nrf, the “fast” delay needs to be used with a 2 cycle delay. With a 48MHz KL27 core clock, one cycle for the FGPIO access and 2 cycle fast delay, this results in a 8MHz SWCLK speed ((48MHz/2)/(1+2)=8MHz). The macro DELAY_FAST_CYCLES=2U was added to the microbit yaml file.
>

microbit-carlos (commented):
Added a couple of clarifications/notes and questions we had.

flit:
Thanks for the PR! I probably won't be able to get to it until this weekend at the earliest.

source/usb/usb_lib.c
```diff
@@ -1065,6 +1065,8 @@ void USBD_Reset_Event(void)
 #endif
 #endif  /* ((USBD_CDC_ACM_ENABLE)) */
 
+__weak void board_usb_sof_event(void) {}
```
mathias-arm, line 1068: That should be __WEAK now

source/board/microbitv2/microbitv2.c
```diff
@@ -52,3 +621,421 @@ const board_info_t g_board_info = {
...
+            // Check first is config is already present in flash
+            // If differences are found, erase and write new config
+            if (0 != memcmp(&gflashConfig, (void *)FLASH_CONFIG_ADDRESS, sizeof(flashConfig_t))) {
+                status = erase_storage_sector(FLASH_CONFIG_ADDRESS);
+
+                if (status != 0) {
+                    pI2cCommand->cmdId = gFlashError_c;
+                }
+                else {
+                    status = program_storage_page(FLASH_CONFIG_ADDRESS, sizeof(flashConfig_t), (uint32_t *) &gflashConfig);
```
mbrossard, line 992: I get an error on this line with gcc: ```
source/board/microbitv2/microbitv2.c:992:21: error: converting a packed 'flashConfig_t' {aka 'struct flashConfig_tag'} pointer (alignment 1) to a 'uint32_t' {aka 'long unsigned int'} pointer (alignment 4) may result in an unaligned pointer value [-Werror=address-of-packed-member]
  992 |                     status = program_storage_page(FLASH_CONFIG_ADDRESS, sizeof(flashConfig_t), (uint32_t *) &gflashConfig);
      |                     ^~~~~~
In file included from source/board/microbitv2/microbitv2.c:35:
source/board/microbitv2/i2c_commands.h:117:25: note: defined here
  117 | typedef __PACKED_STRUCT flashConfig_tag {
      |                         ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

Is there something to fix, or should I disable the warning?

File: source/usb/usb_lib.c (hunk 4 above)
Comment (gerargz) [reply source/usb/usb_lib.c:1068]: I just tested this and the build fails when changing this to __WEAK without including "cmsis_compiler.h". Should I leave it as __weak instead in this PR?

Comment (mathias-arm) [reply source/usb/usb_lib.c:1068]: I am sorry, this will only have to be done after #701 is merged.

File: records/board/kl27z_microbit_bl.yaml [hunk 6]
```diff
@@ -1,4 +1,6 @@
 common:
+    macros:
+        - MICROBIT_LOCK_BOOTLOADER
```
gerargz, line 3: FYI, @mbrossard. The MICROBIT_LOCK_BOOTLOADER symbol was on its own "common:" section in the yaml file and somehow was ignored by progen I think. I moved it to the existing "common:" section in the yaml file and now the symbol was seen correctly by the preprocessor.

source/daplink/interface/swd_host.c
```diff
@@ -1051,6 +1051,19 @@ uint8_t swd_set_target_state_sw(target_state_t state)
...
+            if (!swd_write_dp(DP_CTRL_STAT, 0x0)) {
+                return 0;
+            }
+            
+            // Wait until ACK is deasserted
+            do {
+                if (!swd_read_dp(DP_CTRL_STAT, &val)) {
+                    return 0;
+                }
+            } while ((val & (CSYSPWRUPACK | CDBGPWRUPACK)) == 1);
```
flit, line 1065: ADIv6 says that we must not clear `CSYSPWRUPREQ` and `CDBGPWRUPREQ` at the same time.
ADIv5 says `CSYSPWRUPREQ` must not be set to 1 while `CDBGPWRUPREQ` is set to 0.

While DAPLink doesn't support any targets that use ADIv6 yet, it will happen soon with the Cortex-M55. So it would be best to go ahead and make this compatible by clearing `CSYSPOWERUPREQ` first, followed by `CDBGPWRUPREQ`.

source/hic_hal/freescale/kl27z/DAP_config.h
```diff
@@ -68,7 +68,7 @@ Provides definitions about:
 /// Default communication speed on the Debug Access Port for SWD and JTAG mode.
 /// Used to initialize the default SWD/JTAG clock frequency.
 /// The command \ref DAP_SWJ_Clock can be used to overwrite this default setting.
-#define DAP_DEFAULT_SWJ_CLOCK   5000000         ///< Default SWD/JTAG clock frequency in Hz.
+#define DAP_DEFAULT_SWJ_CLOCK   8000000         ///< Default SWD/JTAG clock frequency in Hz.
```
flit, line 71: I'm wondering if making the default clock 8 MHz is ok. On the micro:bit you have short SWD traces and a known design, so it's always fine. But for other boards it may not be reliable.

If we do decide to change it, it should be configurable with a macro. So this line would need to be wrapped with `#if !defined(DAP_DEFAULT_SWJ_CLOCK)`.

source/hic_hal/freescale/kl27z/usb_config.c
```diff
@@ -68,7 +68,7 @@
 //       <i> when the device is fully operational (bMaxPower)
 //   </h>
 #define USBD_CFGDESC_BMATTRIBUTES   0x80
-#define USBD_CFGDESC_BMAXPOWER      0xFA
+#define USBD_CFGDESC_BMAXPOWER      0x96    // Temporal modification to 300mA for microbit
```
flit, line 71: Probably should make this configurable with an optional `BOARD_USB_BMAXPOWER` macro. The other `usb_config.c` files should be adjusted too.

(We really need to add a docs file that lists all the per-HIC and per-board customization points like weak functions, global variables, and various macros.)

flit (changes requested):
My apologies for taking so long to review! There are a couple things on my list to come back to look at again. Overall, really good work! My comments are mostly nitpicks. 😉

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -124,9 +130,14 @@ error_t flash_decoder_get_flash(flash_decoder_type_t type, uint32_t addr, bool a
...
-            flash_start_local = DAPLINK_ROM_IF_START;
-            flash_intf_local = flash_intf_iap_protected;
+            if (addr_valid && (DAPLINK_ROM_IF_START != addr)) {
+                // Address is wrong so display error message
+                status = ERROR_FD_INTF_UPDT_ADDR_WRONG;
+            } else {
+                // "Target" update in this case would be a 3rd party interface application
+                flash_start_local = DAPLINK_ROM_IF_START;
+                flash_intf_local = flash_intf_iap_protected;
+            }
```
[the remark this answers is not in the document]
microbit-carlos: @flit just to confirm that you are happy with this change affecting all ports?

source/daplink/cmsis-dap/DAP.c
mbrossard: This is a file we keep in sync with CMSIS-DAP.

- Will this change be submitted to CMSIS-DAP?
- If we are to be keeping this change compared to upstream for some time, I would rather keep the new lines where the old ones are removed.

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -110,11 +112,13 @@ static const magic_file_info_t s_magic_file_info[] = {
...
         { "MSD_ON  CFG", kMSDOnConfigFile           },
         { "MSD_OFF CFG", kMSDOffConfigFile          },
+        { "COMP_ON CFG", kImageCheckOnConfigFile    },
+        { "COMP_OFFCFG", kImageCheckOffConfigFile   },
         { "PAGE_ON ACT", kPageEraseActionFile       },
         { "PAGE_OFFACT", kChipEraseActionFile       },
     };
 
-static uint8_t file_buffer[VFS_SECTOR_SIZE];
+static uint8_t file_buffer[VFS_SECTOR_SIZE+100];
```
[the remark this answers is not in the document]
mbrossard: We discussed in #719 removing information to fit in the allocated space (and keep files one sector). This PR actually adds another line. If the state of all those settings need to be exposed, we should think of more compact ways of conveying that information. As a point of comparison we have `USB Interfaces: MSD, CDC, HID, WebUSB`, when we could have had separate more user-friendly (`USB Drag & drop flashing: 1`, `USB CDC UART: 1`, etc.). I feel this verbosity comes at the expense of other interesting information we could have that is harder to look up (project name, HIC name).

I would like `file_buffer` to be eventually removed. Unless we can find another buffer we can re-use in the same way, like @flit said, we would need to refactor file "rendering" to be able to work without buffer (which currently serves only for length computation).

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -334,12 +353,22 @@ static uint32_t read_file_mbed_htm(uint32_t sector_offset, uint8_t *data, uint32
...
+    
+    if (size - VFS_SECTOR_SIZE * sector_offset > VFS_SECTOR_SIZE) {
+        memcpy(data, file_buffer + sector_offset * VFS_SECTOR_SIZE, VFS_SECTOR_SIZE);
+    }
+    else {
+        memcpy(data, file_buffer + sector_offset * VFS_SECTOR_SIZE, size - VFS_SECTOR_SIZE * sector_offset);
+    }
 
-    return update_details_txt_file(data, VFS_SECTOR_SIZE);
+    return size;
```
mathias-arm, line 371: The return value should probably be count of bytes copied.

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -334,12 +353,22 @@ static uint32_t read_file_mbed_htm(uint32_t sector_offset, uint8_t *data, uint32
 // File callback to be used with vfs_add_file to return file contents
 static uint32_t read_file_details_txt(uint32_t sector_offset, uint8_t *data, uint32_t num_sectors)
 {
+    uint32_t size = 0;
 
-    if (sector_offset != 0) {
+    if (sector_offset > 1) {
```
mathias-arm, line 358: This should probably depend on the value of `BOARD_EXTRA_BUFFER`

records/board/microbitv2.yaml
```diff
@@ -4,9 +4,13 @@ common:
         - CDC_LED_DEF=GPIO_LED_ON
         - MSC_LED_DEF=GPIO_LED_ON
         - USB_PROD_STR="BBC micro:bit CMSIS-DAP"
+        - IO_CONFIG_OVERRIDE
+        - DELAY_FAST_CYCLES=2U
```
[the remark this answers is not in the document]
gerargz: Based on the single cycle access of the GPIO in the KL27, a "fast delay" of 2 cycles is needed for reaching an 8MHz SWD clock speed.

source/board/microbitv2/i2c.c
```diff
@@ -0,0 +1,239 @@
+/**
```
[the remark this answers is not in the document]
gerargz: Renamed to "i2c_commands.c".

source/daplink/interface/swd_host.c
gerargz: This change is not implemented yet as it could affect the power consumption, so it should be validated before adding it to this PR or change it in a different PR.

source/hic_hal/freescale/kl27z/DAP_config.h
gerargz: Made this change board specific in the .yaml file.

source/hic_hal/freescale/kl27z/usb_config.c
gerargz: Made this change board specific in the .yaml file.

source/board/microbitv2/microbitv2.c
gerargz: Added comment at the top of the enum in error.h.

source/daplink/drag-n-drop/vfs_user.c
mathias-arm: Never mind. Ignore the previous comment.

source/daplink/interface/swd_host.c
flit: When you split the clearing of these bits, please also add a comment stating that ADIv6 specifies that it must be done like that and to please not merge them back together. 😄

source/daplink/error.h
```diff
@@ -66,6 +66,7 @@ typedef enum {
     ERROR_FD_BL_UPDT_ADDR_WRONG,
     ERROR_FD_INTF_UPDT_ADDR_WRONG,
     ERROR_FD_UNSUPPORTED_UPDATE,
+    ERROR_FD_INCOMPATIBLE_IMAGE,
```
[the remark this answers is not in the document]
microbit-carlos: Added as par of https://github.com/ARMmbed/DAPLink/pull/763#discussion_r692302432

records/board/microbitv2.yaml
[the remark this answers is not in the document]
microbit-carlos: It was discussed to add a comment here explaining this value trigger the else path from DAP.c

source/daplink/drag-n-drop/vfs_user.c
gerargz: Updated calculation.

source/daplink/interface/swd_host.c
gerargz: I have updated it as per ADIv6. Could you review it?

source/daplink/drag-n-drop/flash_decoder.c
```diff
@@ -179,6 +190,28 @@ error_t flash_decoder_get_flash(flash_decoder_type_t type, uint32_t addr, bool a
...
+    error_t status = ERROR_SUCCESS;
+
+    if (daplink_is_interface()) {
+        if (FLASH_DECODER_TYPE_TARGET == type) {
+            if (g_board_info.target_cfg) {
+                if (board_detect_incompatible_image(data, size)){
+                    status = ERROR_FD_INCOMPATIBLE_IMAGE;
+                    flash_intf_target->init();
+                    flash_intf_target->erase_chip();
+                    flash_intf_target->uninit();
```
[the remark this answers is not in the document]
gerargz: Removed 👍

mbrossard (approved):

source/board/microbitv2/pwr_mon.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum {
+    PWR_SOURCE_NONE = 0,
+    PWR_USB_ONLY,
+    PWR_BATT_ONLY,
+    PWR_USB_AND_BATT
```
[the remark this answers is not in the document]
flit: This was never addressed, but at this point let's deal with it after the merge.

mathias-arm (approved):

source/daplink/interface/main.c
```diff
@@ -104,6 +105,43 @@ static main_led_state_t msc_led_state = MAIN_LED_FLASH;
...
+        // Reset button released
+        target_set_state(RESET_RUN);
+        reset_pressed = 0;
+    }
+}
+
+__WEAK void board_handle_powerdown()
+{
+    // TODO: put the interface chip in sleep mode
+}
```
flit, line 138: Just noticed that the `while(1);` from `main()` is missing here.

It's probably ok if the main loop keeps running, and might actually be better: RTX will block on the `osThreadFlagsWait()` call and enter the idle thread which will WFI. 

But has this been tested on a build without the micro:bit override of this hook?

flit (approved):
Approved. Really fantastic work @gerargz!

The only outstanding question is the new one about the missing `while(1);` from the powerdown hook. Ideally I'd like to get an answer before this is actually merged, but it's not a big deal either way.

source/board/microbitv2/microbitv2.c
flit: Was this addressed?

source/daplink/interface/main.c
gerargz: If I remember correctly I initially had the while(1); in the weak function, but this had strange behavior where the code wasn't executing correctly in the "strong" custom function. I think it had to do with how the compiler optimizations were working. Something about the optimizations being done per file, without knowing that there was a "strong" function definition elsewhere. That and the combination of having an endless loop in the weak function, made the optimizations assume wrong things.

I don't see any build that uses the main_powerdown_event() to trigger this event, so the while(1) was probably a placeholder and there's no impact by removing it, I think.

source/board/microbitv2/microbitv2.c
gerargz: Yes, in de84e4c.

source/board/microbitv2/pwr_mon.h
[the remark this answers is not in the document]
gerargz: I missed this one. I'll make a note to raise a PR for this after the merge.

source/daplink/interface/main.c
flit: Ok, cool. Nothing to worry about then. Thanks for the quick answer!

source/board/microbitv2/microbitv2.c
flit: Perfect, thanks!

flit:
All questions addressed, ready for merge! 🚀

PR #771, opened 2021-01-27
Add GitHub Actions on experimental_compilers branch

mbrossard:
GitHub Actions can be used on the `experimental_compilers` with GCC. Enabling Linux workflow on all pushes to the branch. Windows is on-demand because it's slower (in part because it does not support `ccache`) and most changes are not OS-dependent.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,14 +58,14 @@ jobs:
     - name: Install dependencies
       run:  |
         sudo apt install ccache ninja-build
-        for i in ${{ runner.temp }}/arm-gcc/bin/* ; ln -s /usr/bin/ccache /usr/lib/ccache/$(basename $i); done
+        for i in ${{ runner.temp }}/arm-gcc/bin/* ; do ln -s /usr/bin/ccache /usr/lib/ccache/$(basename $i); done
         ccache -M 1Ti
 
     - name: Compile
       run:  |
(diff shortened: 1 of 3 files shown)


cederom:
Some time ago I have invented such CI that would build DAPLink firmware images on pull requests so people would have their pull requests automatically verified and compiled for real world testing. That was important especially when Keil license was needed to build the firmware and was kind of blocker for public DAPLink development. Nothing happened with this idea though.

GCC support  is a giant leap and critical enabler for public free and open development of DAPLink. People now can modify and compile the source code on their own local Open-Source workstations such as FreeBSD or Linux, perform local testing, then create a verified pull requests :-)

These GitHub Actions will build the master branch and generate a distribution packages right? Can they also build Pull Requests when trigerred with some sort of command?

mbrossard:
You can have a look at the [Actions](https://github.com/mbrossard/DAPLink/actions) tab on my fork to see what it currently looks like.

For now this is targeted at the `experimental_compilers` branch (the only that supports GCC). I have restricted for now to manual trigger and on pushes to this branch. We will see what works because there are some GitHub Actions restrictions until workflows are present in the main development branch. We will possibly not be able to enable all scenarios until the branch is merged into `master`.

As a side note, the current CI system tests with hardware. This is quite important but this introduces some reliability and security concerns, which is why it's not run automatically on all PRs.

PR #674, opened 2019-11-18
add a new hic STM32F072CB

dotnfc:
new hic using STM32F072CB chip, it is an ideal daplink interface for small project.

this request is a replacement of [#495](https://github.com/ARMmbed/DAPLink/pull/495)

PR #782, opened 2021-02-17
K22F and KW38-ER-RD board

povalac:
Hello team,

I would like to add a new HIC device K22F and KW38-ER-RD board from NXP to the repo. Can you please check it?

Thank you in advance.
Best regards,
Karel Povalac


mathias-arm:
Similarly to #674. We would prefer to merge new HICs after support for GCC is merged. Would you be able to rebase that PR on `experimental_compilers`?

mbrossard:
I rebased on `experimental_compilers` and did some changes to support GCC. I split the K22F part into [mbrossard:feature/k22f_hic](https://github.com/mbrossard/DAPLink/tree/feature/k22f_hic), and the KW38 interfaces in [mbrossard:feature/k22f_kw38_if](https://github.com/mbrossard/DAPLink/tree/feature/k22f_kw38_if). I was able to do some smoke testing with a FRDM-K22F (it seems with current configuration the HIC thinks the reset button is pushed, so it will always jump into bootloader).

I have a few questions:
- Why did you change the HIC ID?
- It seems the bootloader also has some compilation time assert to check the bootloader and interface sizes are 32 KB and 96 KB respectively. So I made the changes in `daplink_addr.h` to limit Flash to 128 KB, and limited RAM to 24 KB. Those changes make it possible to target MCUs from the K22_100 sub-family (like MK22FN128*10) in addition to the ones in K22_120 (like MK22FN512*12). Was this your intent also when you chose the bl/if sizes?
- There is a K22_50 sub-family and I was wondering if this is why you chose to use 48 MHz clock?
- What is the reason for using `#if defined(IINTERFACE_K22F)` in `source/hic_hal/freescale/k22f/gpio.c`? It seems to me the `INTERFACE_K22F` is always set (for both bootloader and interface).

Without the proper hardware it is not possible for me to fully test. If you can test and fix on top of the `feature/k22f_hic` and update your PR to target `experimental_compilers`. I would prefer to keep the PR for HIC and boards separate.

Issue #768, opened 2021-01-20
USB not recognized


BibyThm:
Hi, I have customized the daplink files for STM32F373CC. The project was compiled using Keil. When it was loaded to the board, the USB was not recognized. It shows "USB is not recognizing the device". Any suggestion to the above problem?

mbrossard:
As mentioned in [PR #750](https://github.com/ARMmbed/DAPLink/pull/750#issuecomment-767964899) to @Ashleyvk, you seem to be working on the same MCU in two different environments. Except for the startup part, you should be able to help each other.

BibyThm:
Any suggestions to the above problem?

Ashleyvk:
I'm facing the same issue too. Has it got something to do with the usbd file for the controller, under the stm32 folder?

Ashleyvk:
![usberror](https://user-images.githubusercontent.com/74317336/111953487-b48dea80-8b0c-11eb-96a2-ac4028dcc47d.JPG)
When the board is connected, it shows that device descriptor failed. The result of USB analyzer is as attached. Any suggestions on this?

symovs:
We are using STM32CubeProgrammer to load the DAPLINK binary in to our custom HIC implemented using STM32F373CC. We are loading the GCC compiled bin file (*_if_crc.bin) with the start address set to 0x0800000 in the STM32CubeProgrammer. 

Are we doing this right ? 
Do we need any DAPLINK specific bootloader for STM32 HIC ? 

@mbrossard @flit thoughts are welcome.

PR #257, opened 2017-04-16
Stm32f103xb interface

c1728p9:
Add support for the stm32f103 interface to DAPLink. Thanks, @zoujixing for doing most of the porting work in PR #133. This patch is based on yours but replaces the HAL with the stm32 cube driver so the license is open source. It also replaces the USB driver with an open source version. Finally, it is rebased and configured for NUCLEO boards.

PR #789, opened 2021-03-08
Add support for stm32f072cb as new HIC

shiva-pandit:
Here is an attempt to add support for `stm32f072cb` as new HIC. Please review and let me know if I missed anything. 

Tested all basic HIC operations on our custom board. Binaries were built using `gcc` toolchain - `gnu-arm-embedded/gcc-arm-none-eabi-9-2020-q2-update`. Have not tested any other toolchain.

- STM32F072CB comes up in DFU mode by default. I was able to flash the `daplink_if` firmware using the `dfu-util`. 
- Using STM32L452 as the target processor, was able to flash and debug an application on target mcu using `openocd/gdb` combo. 
- Verified the serial output from the target mcu on ACM interface of the HIC.

Some of this work is based off of this old PR: https://github.com/ARMmbed/DAPLink/pull/674


mbrossard:
Excellent. I was able to compile this on gcc, armclang and armcc.

I was looking at #674 from @dotnfc and I noticed your version has a different IO configuration. Are you using standard board (like NUCLEO-F072RB, 32F072BDISCOVERY) or your own design?

shiva-pandit:
> Excellent. I was able to compile this on gcc, armclang and armcc.
> 
> I was looking at #674 from @dotnfc and I noticed your version has a different IO configuration. Are you using standard board (like NUCLEO-F072RB, 32F072BDISCOVERY) or your own design?

This is our own design. Unfortunately, I don't have access to any other F072 based boards...

tarokajya:
@mbrossard :It seems that IO configuration #674 from @dotnfc is based on JLINK OB stm32f072.

dotnfc:
Yes, sure. the  IO configuration from  is #674 is based on the [JLINK-OB F072](https://www.segger.com/downloads/jlink/UM08024_JLinkOBSTM32F072.pdf).

symovs:
hi @shiva-pandit, we (@BibyThm, @Ashleyvk and me) were trying to add a new HIC which uses STM32F373CC. While we had followed mostly the https://github.com/ARMmbed/DAPLink/pull/257 for our work we have not been able to connect the HIC PCB we have to our WIN 10 PC and get it enumerated yet as we noted here - https://github.com/ARMmbed/DAPLink/issues/768. After trying out few things which are listed below

1. We are testing on DAPLINK customized for STM32F373CC based HIC.  The STM32F373CC has external pull up resistor to 3.3V and the device is bus-powered. We have tested the hardware with a libusb based CDC device firmware generated from STM32 Cube IDE. So hardware seems alright. 
2. We are testing on Windows 10 and so DAPLINK should work with native windows inbox USB drivers. So doesn't seem to be a driver issue.
3. We have compiled our code using KEIL and GCC and seeing the same issue with both. So definitely not a compiler issue. 
4. As some boards were using custom bootloader firmware to load DAPLINK we checked whether loading with STCubeprogrammer over ST DFU will cause an issue. After some thought, we think we are doing this right by programming the device starting from the address 0x0800000 in the STM32CubeProgrammer
5. Out of the four output files produced by the KEIL compiler - *_if.bin, *_if_crc.bin, *_if_crc_legacy_0x5000.bin, *_if_crc_legacy_0x8000.bin verified that we think we are using the correct file for programming which is the  - *_if.bin / 

Now we are thinking we may have got it wrong with the usbd_STM32F373.c we created. It would be very helpful if you can let us know how the file is generated like in your case this one - source/hic_hal/stm32/stm32f072cb/usbd_STM32F072.c. We are not able to see such a file name in our STM32Cube IDE or MX generated files. We merely copied over the https://github.com/ARMmbed/DAPLink/blob/master/source/hic_hal/stm32/stm32f103xb/usbd_STM32F103.c to our new stm32f373xc folder and renamed the file to usbd_STM32F373.c

@brianesquilona @flit @mbrossard @dotnfc your inputs if any would also be greatly appreciated.

mathias-arm:
@shiva-pandit: it seems the I/O configuration suggested by @dotnfc is popular so I think it would be better to have it as the default. Could you update `IO_Config.h` to match that configuration (see #674 and the document linked above). You should be able to easily use the `IO_CONFIG_OVERRIDE` mechanism for your board.

I am getting a [32F072BDISCOVERY](https://www.st.com/en/evaluation-tools/32f072bdiscovery.html) to test this.

mathias-arm:
I looked at the different pinouts and the situation is a little bit more complicated because this version uses two pins for SWDIO. If you are still interested in this work, I can work with you (in [mbrossard/DAPLink:feature/stm32f072cb_hic](https://github.com/mbrossard/DAPLink/tree/feature/stm32f072cb_hic)) on modifying `IO_Config.h` and `DAP_config.h` so both setups can be supported, otherwise we can simply close this PR.

shiva-pandit:
@mathias-arm we can close this PR. support for discovery board should be good enough. thanks.

mathias-arm:
Thank you.

PR #794, opened 2021-03-20
Add software auto reset to Atmel and stm32

Kevincoooool:
The target chip can auto reset after download with this pr


mathias-arm:
This looks like a layering violation. You should look at why `target_set_state(RESET_PROGRAM)` (or possibly `swd_set_target_state_hw(RESET_PROGRAM)`) does not work in your case. Not sure if there is a generic (or HIC-specific) way to detect that hardware reset does not work and fall back to software reset.

cederom:
There was an idea to unify reset once https://github.com/ARMmbed/DAPLink/pull/371

Kevincoooool:
Hello, because my hardware does not set aside RST pin for DAPLINK, so I can only use SWD to perform a software reset. This PR has been tested to work normally and there is no other impact

PR #799, opened 2021-03-31
Add custom reset for Maxim targets

khariya:


records/board/max32620fthr.yaml
```diff
@@ -4,3 +4,4 @@ common:
             - source/board/max32620fthr.c
         family:
             - source/family/maxim/max32620/target.c
+            - source/family/maxim/max32620/target_reset_max32620.c
```
flit, line 7: This file is missing a newline at the end (the red icon shown by GitHub).

flit (changes requested):
Very nice! The only request (aside from the one other comment) is to please change the copyright to Maxim's for the new files. You might want to correct the copyright on existing files from Maxim, or at least add a Maxim copyright, but I don't think it's mandatory (should have been doing this all along).

mathias-arm:
Looking at #845 adding a third "Family ID", it seems to me that the  same `kMaxim_MAX326xx_FamilyID`, `g_maxim_max326xx` and `target_set_state_max326xx()` could be used for max32620, max32625 and  max32660 (possibly max32665?).

mathias-arm:
I created a rebased branch of both PR in [mathias-arm:fix/maxim-reset-main](https://github.com/mathias-arm/DAPLink/tree/fix/maxim-reset-main). I merged the max32620 and max32630 family, but noticed there is a small difference for max32660 (the flash controller bit is absent and its address is different anyways).

I am wondering if max32625 and max32665 targets should be modified as well.

Let me know what you think.

mathias-arm:
Can you rebase so I can merge it?

mathias-arm:
Merged in [4a3a9807]

PR #803, opened 2021-04-21
Add LPC55xx HIC support to experimental_compilers

mbrossard:
@flit is the main author, tested with [MCU-Link](https://www.nxp.com/design/development-boards/lpcxpresso-boards/mcu-link-debug-probe:MCU-LINK).


records/rtos/rtos-cm33.yaml
```diff
@@ -0,0 +1,36 @@
+common:
+    macros:
+        - OS_TICK=10000
```
mathias-arm, line 3: For `rtos2` this should be OS_TICK_FREQ=100

PR #804, opened 2021-04-25
Cleanup: usb_config.c and usb_buf.h

mbrossard:
I was trying to understand the changes between the versions of `usb_config.c` (one for each HIC) and it lead me down to a path to remove the gratuitous differences. This also highlighted a couple of minor issues.

After those changes the sets `k20dx/kl26z/kl27z`, `k26f/lpc4322/lpc55xx` and `max32620/max32625` share identical files.


mbrossard:
It seems that `USBD_DEVDESC_BCDDEVICE` is defined to `0x1000` or `0x0100` whether the bulk endpoints are enabled or not. If true this should be documented and possibly automated (by relying on the definition of BULK_ENDPOINT).

asier70:
I think that my issue:
[https://github.com/ARMmbed/DAPLink/issues/810#issue-893226621]
should be placed in this conversation because it regards to all usb_config.c files.

Issue #727, opened 2020-08-26
MAX32625PICO not listed here


Gerriko:
I went to https://armmbed.github.io/DAPLink/ in search of the latest DAPLink firmware for the MAX32625PICO.

The only board found was MAX32625, which does not work.

Through trial and error I discovered that the MAX32625PICO bin files are in fact stored elsewhere in the MBED.com labyrinth, which is a not very useful location (IMHO).

https://os.mbed.com/teams/MaximIntegrated/wiki/MAX32625PICO-Firmware-Updates

Maybe you could update the github DAPLink search facility accordingly.

0xc0170:
Thanks for reporting.

My understanding is that daplink page searches only the releases from this repository (aka official releases). There are many more (even from private forks) available on os.mbed.com

mathias-arm:
The images are now at https://github.com/MaximIntegratedMicros/max32625pico-firmware-images

ozersa:
> uld update the github DAPLink search facility accordingly.

Update, it is here:
https://github.com/MaximIntegrated/max32625pico-firmware-images/

Issue #711, opened 2020-05-01
LPCXpresso55S69 not listed at https://armmbed.github.io/DAPLink/


maclobdell:
LPCXpresso55S69 is supported in the last few releases of DAPLink, but when we go to the DAPLink releases page, it does not show up.

toyowata:
@maclobdell 

The LPCXpresso55S69 binary is available here:

https://github.com/ARMmbed/DAPLink/blob/gh-pages/firmware/0253_lpc4322_lpc55S69xpresso_0x10000.bin

However, the target is not listed here:
https://github.com/ARMmbed/DAPLink/blob/gh-pages/_data/update.yml

You can probably make pull-request to the gh-pages branch?

maclobdell:
@0xc0170 - what is the normal process for this? Is this something that was maintained by the old maintainers? For every new board, do we need to ask partners to update the gh-pages branch?

0xc0170:
I dont know the process but looking at the history, we maintain the pages.

@maclobdell would you be able to send a PR adding this target to the yml file?

Thanks @toyowata for the references

flit:
The internal release procedure on Confluence and OneNote documents the process, I believe.

We need to do a new release, anyway. It's been a long time. Note that we need to skip version 0255, as this is reserved for micro:bit.

PR #820, opened 2021-06-07
Update DAPLink site to include 0254 and past releases

mbrossard:
The result can be seen at https://mbrossard.github.io/DAPLink/ (no web designer, but it turned out OK, I think)

- I added a selection control for the version (if there is more than one)
- I changed the format for the release YAML files to deduplicate the instructions.
- I added the script used for converting from the old format (`tools/convert.py`). There is another script (`tools/update.py`) to update `_data/firmwares.yml` based on the versions that exist in `firmware`. I need to make some additions to the tool and document to make adding new releases even simpler. There are also pending changes on the release generation side.
- All previous versions were already in the repository, so I just undeleted them.

Addresses #727 and #711.


mathias-arm:
We should probably pull the plug on Travis for the `gh-pages` branch and replace with some GitHub Actions.

PR #836, opened 2021-08-06
Changes to DETAILS.TXT (fix for #719)

mbrossard:
As discussed in #719 some changes to `DETAILS.TXT` to keep it under 512 bytes. Using the output of `git describe HEAD` instead of simply short commit ID.

For `kl27z_microbit_if` this results in 483 bytes file:
```
# DAPLink Firmware - see https://mbed.com/daplink
Build ID: alpha5-14-gba49a0b8 (gcc)
Unique ID: 9903360258994e45001290120000005c000000009796990b
HIC ID: 9796990b
Auto Reset: 1
Automation allowed: 1
Overflow detection: 0
Page erasing: 0
Daplink Mode: Interface
Interface Version: 0256
Bootloader Version: 0255
USB Interfaces: MSD, CDC, HID, WebUSB
Bootloader CRC: 0x828c6069
Interface CRC: 0xd11fb543
Remount count: 0
URL: https://microbit.org/device/?id=9903&v=0256
```

### DIFF
--- a/source/daplink/drag-n-drop/vfs_user.c
+++ b/source/daplink/drag-n-drop/vfs_user.c
@@ -4,7 +4,7 @@
  *
  * DAPLink Interface Firmware
  * Copyright (c) 2009-2020, ARM Limited, All Rights Reserved
- * Copyright 2019, Cypress Semiconductor Corporation 
+ * Copyright 2019, Cypress Semiconductor Corporation
  * or a subsidiary of Cypress Semiconductor Corporation.
  * SPDX-License-Identifier: Apache-2.0
  *
@@ -497,6 +497,20 @@ static uint32_t update_html_file(uint8_t *data, uint32_t datasize)
     return expand_info(data, datasize);
 }
 
+#if defined(__CC_ARM)
+#define COMPILER_DESCRIPTION "armcc"
+#elif (defined(__ARMCC_VERSION) && (__ARMCC_VERSION >= 6010050))
+#define COMPILER_DESCRIPTION "armclang"
+#elif defined(__GNUC__)
+#define COMPILER_DESCRIPTION "gcc"
+#endif
+
+#if (GIT_LOCAL_MODS)
+#define LOCAL_MODS ", local mods"
+#else
+#define LOCAL_MODS ""
+#endif
+
 static uint32_t update_details_txt_file(uint8_t *data, uint32_t datasize)
 {
     uint32_t pos=0;
@@ -504,10 +518,13 @@ static uint32_t update_details_txt_file(uint8_t *data, uint32_t datasize)
 
     char *buf = (char *)data;
 
-    //Needed by expand_info strlen
+    // Needed by expand_info strlen
     memset(buf, 0, datasize);
 
     pos += util_write_string(buf + pos, "# DAPLink Firmware - see https://mbed.com/daplink\r\n");
+    // Build ID
+    pos += util_write_string(buf + pos, "Build ID: " GIT_DESCRIPTION \
+        " (" COMPILER_DESCRIPTION LOCAL_MODS ")\r\n");
     // Unique ID
     pos += util_write_string(buf + pos, "Unique ID: @U\r\n");
     // HIC ID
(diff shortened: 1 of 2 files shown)


microbit-carlos:
Thanks Mathias!

With the "local mods" text and the micro:bit V2 features `Incompatible image detection: 1` entry we get dangerously close to the 512 limit, at 511 bytes:

```
# DAPLink Firmware - see https://mbed.com/daplink
Build ID: alpha5-14-gba49a0b8 (gcc, local mods)
Unique ID: 9903360258994e45001290120000005c000000009796990b
HIC ID: 9796990b
Auto Reset: 1
Automation allowed: 1
Overflow detection: 0
Incompatible image detection: 1
Page erasing: 0
Daplink Mode: Interface
Interface Version: 0256
Bootloader Version: 0255
USB Interfaces: MSD, CDC, HID, WebUSB
Bootloader CRC: 0x828c6069
Interface CRC: 0xd11fb543
Remount count: 0
URL: https://microbit.org/device/?id=9903&v=0256

```

Are you still considering the removal of `- see https://mbed.com/daplink`?

Also, any reason not to place the `Build ID` line where the `Git SHA` used to be?

source/daplink/drag-n-drop/vfs_user.c
```diff
@@ -517,22 +534,6 @@ static uint32_t update_details_txt_file(uint8_t *data, uint32_t datasize)
...
-#endif
-
-    // GIT sha
-    pos += util_write_string(buf + pos, "Git SHA: ");
-    pos += util_write_string(buf + pos, GIT_COMMIT_SHA);
-    pos += util_write_string(buf + pos, "\r\n");
-    // Local modifications when making the build
-    pos += util_write_string(buf + pos, "Local Mods: ");
-    pos += util_write_uint32(buf + pos, GIT_LOCAL_MODS);
-    pos += util_write_string(buf + pos, "\r\n");
```
c1728p9, line 535: As an alternative to compressing these fields to fit into 512 bytes, what are your thoughts on updating this function to support arbitrary length details.txt files? One possible way to do this would be to let DETAILS.TXT be written in chunks, so for each sector that needs to be read it is called with a different offset.

Example implementation of update_details_txt_file to allow read from arbitrary position:
```c
static uint32_t update_details_txt_file(uint32_t pos, uint8_t *buf, uint32_t datasize) {

    ...

    pos += util_write_string_in_region(buf, datasize, buf + pos, "Automation allowed: ");
    pos += util_write_string_in_region(buf, datasize, buf + pos, config_get_automation_allowed() ? "1" : "0");
    pos += util_write_string_in_region(buf, datasize, buf + pos, "\r\n");
    pos += util_write_string_in_region(buf, datasize, buf + pos, "Overflow detection: ");
    pos += util_write_string_in_region(buf, datasize, buf + pos, config_get_overflow_detect() ? "1" : "0");
    pos += util_write_string_in_region(buf, datasize, buf + pos, "\r\n");
    
    ...
}
```

DETAILS.TXT creation (buffer size of 10 rather than 512 used):
```
....Automa                                              ...
          tion allow                                    update_details_txt_file(200, buf, 10);
                    ed: 1\nOve                          update_details_txt_file(210, buf, 10);
                              rflow dete                update_details_txt_file(220, buf, 10);
                                        ction: 0..      ...
```

mbrossard:
> With the "local mods" text and the micro:bit V2 features Incompatible image detection: 1 entry we get dangerously close to the 512 limit, at 511 bytes

Release versions will not have "local mods" and will have a shorter tag name. But yes, there is not a lot of space left.

> Are you still considering the removal of `- see https://mbed.com/daplink`?

The link needs to be changed or removed. There is also the question of whether we remove the `URL:` line.

> Also, any reason not to place the `Build ID` line where the `Git SHA` used to be?

I am trying to push up all strings that could be coalesced into a single `util_write_string()` call and do not require dynamic expansion.

source/daplink/drag-n-drop/vfs_user.c
mbrossard: I was thinking about something similar, but one problem is dealing with the `expand_info()` call. A `util_write_string_expand_in_region()` function would be tricky to implement without an additional buffer (e.g. `@R` is expanded to the URL that can contain other escape sequences like board ID or version).

mbrossard: This would also eventually allow to remove `file_buffer` which is a `VFS_SECTOR_SIZE` (512 bytes) array that exists only to compute the size of the VFS files.

mbrossard:
Also considering replacing `Firmware` from the first line and replacing it with `Bootloader` and `Interface`, and removing `Daplink Mode:` line.

source/daplink/drag-n-drop/vfs_user.c
c1728p9: For anything that needs to be expanded you could probably use a temporary stack buffer for the expansion and then copy the result with `util_write_string_in_region`

c1728p9:
@mbrossard just a heads up, changing some of these fields may effect DAPLink testing and mbedls. You may want to run DAPLink tests and try mbedls if you are going to change/remove fields to ensure nothing breaks.

c1728p9:
@flit I don't now how Mbed uses this. If mbedls picks up the board its probably ok. I more meant the DAPLink tests.

mbrossard:
In my tests, `mbedls` was working.

source/daplink/drag-n-drop/vfs_user.c
mbrossard: I did an implementation of this idea in [3624cad](https://github.com/mbrossard/DAPLink/commit/3624cad4e0b916eeed7e4b14ffa3043a985c645e). It saves 512 bytes of RAM and as such takes ~568 bytes more Flash.

mbrossard:
@c1728p9 / @flit: Should I drop this PR in favor of #839?

mathias-arm:
Superseded by #839

PR #837, opened 2021-08-10
LPC55xx: Fix USART0 to USB CDC ACM

iosabi:
This series of patches fix various aspects of the LPC55xx USART CMSIS driver as used by the `uart.c` to expose the USB serial terminal.

### DIFF
--- a/source/hic_hal/nxp/lpc55xx/LPC55S69/drivers/fsl_usart_cmsis.c
+++ b/source/hic_hal/nxp/lpc55xx/LPC55S69/drivers/fsl_usart_cmsis.c
@@ -683,16 +683,12 @@ static int32_t USART_NonBlockingControl(uint32_t control, uint32_t arg, cmsis_us
     {
         /* Abort Send */
         case ARM_USART_ABORT_SEND:
-            usart->resource->base->FIFOINTENSET &= ~USART_FIFOINTENSET_TXLVL_MASK;
-            usart->handle->txDataSize = 0;
-            usart->handle->txState    = kUSART_TxIdle;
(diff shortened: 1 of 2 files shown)


source/hic_hal/nxp/lpc55xx/LPC55S69/drivers/fsl_usart_cmsis.c
```diff
@@ -683,16 +683,12 @@ static int32_t USART_NonBlockingControl(uint32_t control, uint32_t arg, cmsis_us
...
-            usart->handle->txState    = kUSART_TxIdle;
+            USART_TransferAbortSend(usart->resource->base, usart->handle);
             return ARM_DRIVER_OK;
 
         /* Abort receive */
         case ARM_USART_ABORT_RECEIVE:
-            usart->resource->base->FIFOINTENSET &= ~USART_FIFOINTENSET_RXLVL_MASK;
-            usart->handle->rxDataSize = 0U;
-            usart->handle->rxState    = kUSART_RxIdle;
+            USART_TransferAbortReceive(usart->resource->base, usart->handle);
```
flit, line 691: These change should probably be upstreamed to https://github.com/nxpmicro/mcux-sdk so they aren't at risk of being overwritten if we update the drivers.

iosabi: I'm happy to send this upstream, but I'm not familiar with the mcux-sdk and I don't see the `fsl_usart_cmsis.c` file there. I ran the `west` command to fetch the core code and the modules but I can't find this file. Could you provide a pointer to it?

PR #840, opened 2021-08-14
LPC55xx: Fix uart_write_data() with pending transfers.

iosabi:
`uart_write_data()` uses a circular buffer to enqueue next TX bytes and
a single `cb_buf.tx` byte for the "current" byte it is sending. This
function was always calling `USART_INSTANCE.Send` without checking
whether a transfer was already in progress, which would cause the
fsl_uart_cmsis.c to override the current transfer if there was one in
progress. This would cause the next byte in the circular buffer (the
one returned by `pop`) to be discarded when `uart_write_data()` is
called and there is already data pending in the circular buffer. This
bug can be reliably reproduced when issuing multiple writes on the
/dev/ttyACM0 device on a short period of time.

This patch fixes the problem by adding a flag that tells whether the
current transfer is still ongoing or not. At the moment, since all the
`Send()` calls use a size of 1 this flag can only be 0 or 1, but it will
increased in the future for a more performant implementation.

Tested on Linux sending multiple writes of various lengths and checking
the receiving end on a different USB-to-UART adapter.

### DIFF
--- a/source/hic_hal/nxp/lpc55xx/uart.c
+++ b/source/hic_hal/nxp/lpc55xx/uart.c
@@ -45,6 +45,10 @@ circ_buf_t read_buffer;
 uint8_t read_buffer_data[BUFFER_SIZE];
 
 struct {
+    // Number of bytes pending to be transferred. This is 0 if there is no
+    // ongoing transfer and the uart_handler processed the last transfer.
+    volatile uint32_t tx_size;
+
     uint8_t rx;
     uint8_t tx;
 } cb_buf;
@@ -60,6 +64,7 @@ void clear_buffers(void)
 int32_t uart_initialize(void)
 {
     clear_buffers();
+    cb_buf.tx_size = 0;
     Driver_USART0.Initialize(uart_handler);
     Driver_USART0.PowerControl(ARM_POWER_FULL);
 
@@ -73,6 +78,7 @@ int32_t uart_uninitialize(void)
     Driver_USART0.PowerControl(ARM_POWER_OFF);
     Driver_USART0.Uninitialize();
     clear_buffers();
+    cb_buf.tx_size = 0;
 
     return 1;
 }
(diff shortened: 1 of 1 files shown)


iosabi:
I have [this other work](https://github.com/iosabi/DAPLink/commits/lpc55_uart_tx2) to improve the performance to the TX side, because it is kinda expensive to call Send() with just 1 byte specially a higher baudrates. I wanted to keep to this pull request restricted to just fix the TX issue and then do that other commit to improve on the performance, but I can combine them if you want.

flit:
Thanks again for your work on this!

source/hic_hal/nxp/lpc55xx/uart.c
```diff
@@ -189,15 +195,17 @@ int32_t uart_write_data(uint8_t *data, uint16_t size)
         return 0;
     }
 
-    uint32_t cnt = 0;
-    if (circ_buf_count_used(&write_buffer) > 0) {
+    // Disable interrupts to prevent the uart_handler from modifying the
+    // circular buffer at the same time.
+    NVIC_DisableIRQ(USART_IRQ);
+    uint32_t cnt = circ_buf_write(&write_buffer, data, size);
+    if (cb_buf.tx_size == 0 && circ_buf_count_used(&write_buffer) > 0) {
```
iosabi, line 202: given that the circular buffer has calls to disable/enable interrupts I *think* that I might be able to remove the NVIC_DisableIRQ and NVIC_EnableIRQ calls as this is written (I'm leaning towards I think I can remove them but still need to convince myself). Note that the `circ_buf_count_used(&write_buffer) > 0` part is needed even if we know that we just added >0 bytes to the circular buffer in that case. What do you think?

flit (approved):

source/hic_hal/nxp/lpc55xx/uart.c
flit: Well, ideally yes you could remove the NVIC IRQ disable/enable calls due to the circ_buf interrupt masking. But it looks like there is a race condition in `circ_buf_write()` in how it doesn't wrap the whole call in interrupt masking and instead relies on `circ_buf_push()` protecting each byte insert, with only one check for MIN() and no for loop exit on overflow. So for now it's probably best to keep the IRQ disabling as you have it.

flit: (Added #841 for the general `circ_buf_write()` issue.)

c1728p9:
I ran into this same problem on the nrf52820, which I think had its code copied from the LPC55xx - https://github.com/mathias-arm/DAPLink/pull/4

iosabi:
@c1728p9 you might want to take a look at #842 too then.

PR #842, opened 2021-08-14
LPC55xx: Improve UART TX performance.

iosabi:
Without this patch, transmitting at 921600 baudrate with 8n1 mode yields
to about 49.8KiB/s instead of the theoretical limit of 90 KiB/s. The
reason for this is because between the rising edge of the last data bit
(start of the STOP bit) and the falling edge before the START bit of the
next transfer we observe about 9.7us gap while the minimum is just the
length of a single STOP bit (about 1.1us). This gap is attributed to the
CPU time running the event handler (uart_handler) and scheduling a new
transfer. The performance was measured using a USB-to-UART probe running
with: `picocom -b 921600 /dev/ttyUSB0 | pv >/dev/null` and also looking
at the TX signal on the oscilloscope.

This commit creates two new functions in the circular buffer module that
let you peek into the internal buffer (avoiding a memcpy) and more
importantly allowing to return larger buffers to be passed to the UART
driver in a single Send() call. The UART module in the chip already has
a small FIFO that would allow the non-blocking transfer to enqueue the
buffer with 0 delay between consecutive bytes of the same transfer.
Given the 512 byte circular buffer size and the maximum 128 byte
transfer size this means that only one every 128 bytes would have a
~9us bigger gap with the previous one than needed. At a baudrate of
921600, 128 bytes take 1389 us to transmit so a 9 us delay is about a
1% performance degradation.

After this patch the effective TX bandwidth goes up 89.4 KiB/s, just shy
of the theoretical maximum of 90 KiB/s and consistent with the expected
performance.

### DIFF
--- a/source/daplink/circ_buf.c
+++ b/source/daplink/circ_buf.c
@@ -134,3 +134,47 @@ uint32_t circ_buf_write(circ_buf_t *circ_buf, const uint8_t *data, uint32_t size
 
     return cnt;
 }
+
+const uint8_t* circ_buf_peek(circ_buf_t *circ_buf, uint32_t* size)
+{
+    uint32_t cnt;
+    uint8_t* ret;
+    cortex_int_state_t state;
+
+    state = cortex_int_get_and_disable();
+
+    if (circ_buf->tail >= circ_buf->head) {
+        cnt = circ_buf->tail - circ_buf->head;
+    } else {
+        // We can't peek all the bytes in the circular buffer in this case.
+        cnt = circ_buf->size - circ_buf->head;
+    }
+    ret = circ_buf->buf + circ_buf->head;
+
+    cortex_int_restore(state);
+
+    if (size) {
+        *size = cnt;
+    }
+    return ret;
+}
+
+void circ_buf_pop_n(circ_buf_t *circ_buf, uint32_t n)
+{
+    cortex_int_state_t state;
+
+    state = cortex_int_get_and_disable();
+
+    if (circ_buf->tail >= circ_buf->head) {
+        util_assert(circ_buf->tail - circ_buf->head <= n);
+        circ_buf->head += n;
+    } else {
+        util_assert(circ_buf->tail + circ_buf->size - circ_buf->head <= n);
+        circ_buf->head += n;
+        if (circ_buf->head >= circ_buf->size) {
+            circ_buf->head -= circ_buf->size;
+        }
+    }
+
+    cortex_int_restore(state);
+}
--- a/source/daplink/circ_buf.h
+++ b/source/daplink/circ_buf.h
@@ -57,6 +57,15 @@ uint32_t circ_buf_read(circ_buf_t *circ_buf, uint8_t *data, uint32_t size);
 // Attempt to write size bytes to the buffer. Return the number of bytes written
 uint32_t circ_buf_write(circ_buf_t *circ_buf, const uint8_t *data, uint32_t size);
 
+// Returns a pointer to the next byte on the circular buffer and stores in the
+// value pointed by "size" the number of bytes available in that region, which
+// may be less than the total number of bytes in the circular buffer.
+const uint8_t* circ_buf_peek(circ_buf_t *circ_buf, uint32_t* size);
+
+// Remove n bytes from the front of the circular buffer. The values are
+// discarded.
+void circ_buf_pop_n(circ_buf_t *circ_buf, uint32_t n);
+
 #ifdef __cplusplus
 }
 #endif
--- a/source/hic_hal/nxp/lpc55xx/uart.c
+++ b/source/hic_hal/nxp/lpc55xx/uart.c
@@ -50,7 +50,6 @@ struct {
     volatile uint32_t tx_size;
 
     uint8_t rx;
-    uint8_t tx;
 } cb_buf;
 
 void uart_handler(uint32_t event);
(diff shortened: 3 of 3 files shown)


flit (approved):

mbrossard:
@iosabi, @flit: Running the DAPLink Validation tests with the LPC55S69 HIC, I have encountered assertions at this line: https://github.com/ARMmbed/DAPLink/blob/develop/source/daplink/circ_buf.c#L169

I will do a partial revert (only on `source/hic_hal/nxp/lpc55xx/uart.c`) to disable this optimization

iosabi:
@mbrossard Thanks for the mention. I'm not familiar with the Validation tests, could you share some details with me to try to reproduce the issue on my end?

mbrossard:
The validation tests are in the `test` folder of DAPLink.

It requires firmware binaries for the target. The original code was authored by @c1728p9 
for MbedOS 2 (https://os.mbed.com/users/c1728p9/code/daplink-validation/).
It has since been ported it to MbedOS 5.x (https://github.com/mbrossard/daplink-validation) and a functional equivalent for Zephyr has been created (https://github.com/mbrossard/zdv). The MbedOS 5.x has binary releases you can simply download.

Since a "board" (interface + target) the test project `lpc55s69_nrf52840dk_test_if` was created to test the lpc55s69. It's a combination of a MCU-Link and a nRF52840-DK board. Connect Cortex-M debug cable between the two, Dupont cables from MCU-Link VCOM to P0.08 and P0.06 connectors on nRF52840-DK, and nRF52840-DK should be powered from second micro-USB (not the debug one).

To build the interface firmware start the tests, run from the DAPLink:
```
(venv) $ python ./tools/progen_compile.py -t make_gcc_arm lpc55s69_nrf52840dk_test_if -v --parallel
[...]
(venv) $ python ./test/run_test.py -t make_gcc_arm --firmware lpc55s69_nrf52840dk_test_if \
    --targetdir ~/Downloads/daplink-validation-v0.0.3/ --quick --logdir lpc55s69_nrf52840dk
```

If you need to create a separate test project, pick a target that is already supported by DAPLink and `daplink-validation`.
In order to get the test to work you will need to edit `test/info.py` to add a line similar to the one I added for the test project:
```
    (   0x1102,     VENDOR_TO_FAMILY('Nordic', 2),      'lpc55s69_nrf52840dk_test_if',              'lpc55s69_bl',      'Nordic-nRF52840-DK'                    ),
```
I re-use the board ID (`0x1102`) for the target, otherwise you will need to patch pyOCD to add the new board ID and map to the proper MCU target. The last element on the line needs to match the names of the target firmware files (`Nordic-nRF52840-DK.bin` and `Nordic-nRF52840-DK.hex`).

mbrossard:
Note: the `lpc55s69_nrf52840dk_test_if` project is part of #927 which has not been merged yet.

iosabi:
@mbrossard I wasn't able to run the tests with that command. I keep getting `No board to test firmware lpc55s69_nrf52840dk_test_if`.

Can you share more information about which test fails / produces the assert? Do you have a stack trace of the assert point?

mbrossard:
> @mbrossard I wasn't able to run the tests with that command. I keep getting `No board to test firmware lpc55s69_nrf52840dk_test_if`.

I am surprised. Once the `lpc55s69_nrf52840dk_test_if` firmware is installed running `pyocd list` should identify the nRF52840-DK and the Unique ID should start with `1102`:

```
(venv) $ pyocd list
  #   Probe                    Unique ID
---------------------------------------------------------------------------------
  0   nRF52840-DK [nrf52840]   11023602a6d1d6915e5c1fe529c58fa238ffe6a64c504355
```

> Can you share more information about which test fails / produces the assert? Do you have a stack trace of the assert point?

It is not clear which of the UART tests causes the issue. It is only the next test series that looks at the filesystem and discovers the `ASSERT.TXT` file. A good candidate might be the "Block test" that sends 8 KiB of dummy data.

iosabi:
@mbrossard It turns out that `mbedls` doesn't show the device at all unless the USB drive it creates is mounted. I managed to run the tests on some combination of boards and configs and reproduced the issue. It was easy to fix once I attached a debug probe to the debug probe and got a stack trace from GDB. I uploaded #932 with a fix. It is possible that I never ran this with asserts enabled (I only ran `lpc55s69_if` compiled with cmake, I don't know if that enables asserts) and I definitively didn't test changing the baudrate right after a large TX. Anyway the assert is always inside an ISR so it doesn't crash the program or remount the VFS, so I wouldn't know about it.

This testing setup is certainly over-complicated. I spent many hours trying to get that setup with a board that's supported by mbed os, daplink project and pyocd and compile this ad-hoc validation-test program that's just an echo. For a serial interface test, I think it makes more sense to have a setup where you either just short RX and TX pins, or connect them to another serial port (like a USB to UART adapter). Having to deal with the target is an unnecessary complication.

mathias-arm:
@iosabi: You are right we could probably write a loopback test UART for discrete probes. Consider though that the whole test suite does test a much larger scope of DAPLink functionality, and in many boards the interface UART is connected only to the target. The MbedOS (or Zephyr now) firmware is used to test SWD functionality, drag-n-drop flashing, etc.
There are many constraints and limitations with the tests, but they exist (I am grateful to their authors), their coverage is more than decent, they work, they are repeatable and they catch bugs.

PR #845, opened 2021-08-23
Add MAX32660EVSYS board

ozersa:
Signed-off-by: Sadik.Ozer <Sadik.Ozer@maximintegrated.com>

### DIFF
--- a/projects.yaml
+++ b/projects.yaml
@@ -654,6 +654,10 @@ projects:
         - *module_if
         - *module_hic_max32625
         - records/board/max32630fthr.yaml
+    max32625_max32660evsys_if:
+        - *module_if
+        - *module_hic_max32625
+        - records/board/max32660evsys.yaml
     max32625_max32665hsp_if:
         - *module_if
         - *module_hic_max32625
--- a/records/board/max32660evsys.yaml
+++ b/records/board/max32660evsys.yaml
@@ -4,3 +4,4 @@ common:
             - source/board/max32660evsys.c
         family:
             - source/family/maxim/max32660/target.c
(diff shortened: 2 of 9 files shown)


mathias-arm:
It seems this PR duplicates functionality added #799. Could you work together to factorize this into one `kMaxim_MAX326xx_FamilyID` family with associated reset?

ozersa:
Thanks for review and idea,
We would like to send a separate PR for that purpose.

mathias-arm:
I created a rebased branch of both PR in [mathias-arm:fix/maxim-reset-main](https://github.com/mathias-arm/DAPLink/tree/fix/maxim-reset-main). I merged the max32620 and max32630 family, but noticed there is a small difference for max32660 (the flash controller bit is absent and its address is different anyways).

I am wondering if max32625 and max32665 targets should be modified as well.

Let me know what you think.

ozersa:
Thanks, it needs for MAX32665 and MAX32625, 
These devices (MAX32620/25/60/65) has different die type so probably there will be same differentiation. 
The purpose of this feature it to able to wakeup target even in low power case.
So that it needs to be tested on hardware, especially when target in deepsleep mode.

ozersa:
Again I prefer to centralize them on separate PR in future.
Thanks

mathias-arm:
Ok I will merge this PR, and will expect other PR for other Maxim targets. Should #799 be closed?

ozersa:
Yes please merge  #799 too.
Thank you.

mathias-arm:
I looked and #799 needs to be rebased to be merged.

PR #851, opened 2021-09-18
Fix for micro:bit universal hex support (fixes #847)

### LINKED ISSUE #847: micro:bit: Universal hex flashing on develop
Based on testing fd42dbddd134e6e6cc1f017368ba4952c8ea180f

Attempting to flash a universal hex (https://tech.microbit.org/software/spec-universal-hex/) such as the attached crashes DAPLink running on a V2.00 board build with GCC

On armcc MSD flashing is 'globally' broken so can't be sure if this problem is also present there.

[uhex-sectors-upy.hex.zip](https://github.com/ARMmbed/DAPLink/files/7138490/uhex-sectors-upy.hex.zip)

cc @gerargz @mathias-arm @microbit-carlos

mbrossard:
- The extraneous store for load_unaligned_record seems to be a compiler issue.
- The exit with HEX_PARSE_OK in the else clause seems very wrong.
  If the version does not match one should simply continue to the next line and
  not mark the whole buffer as processed.
- The shadow_line content was never used only ever written to.

### DIFF
--- a/source/daplink/drag-n-drop/intelhex.c
+++ b/source/daplink/drag-n-drop/intelhex.c
@@ -82,7 +82,7 @@ static uint8_t validate_checksum(hex_line_t *record)
     return (result == 0);
 }
 
-static hex_line_t line = {0}, shadow_line = {0};
+static hex_line_t line = {0};
 static uint32_t next_address_to_write = 0;
 static uint8_t low_nibble = 0, idx = 0, record_processed = 0, load_unaligned_record = 0;
 static uint16_t binary_version = 0;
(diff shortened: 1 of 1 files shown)


source/daplink/drag-n-drop/intelhex.c
```diff
@@ -178,10 +178,6 @@ hexfile_parse_status_t parse_hex_blob(const uint8_t *hex_blob, const uint32_t he
                                             // Save next address to write
                                             next_address_to_write = ((next_address_to_write & 0xffff0000) | line.address) + line.byte_count;
                                         }
-                                        else {
-                                          status = HEX_PARSE_OK;
-                                          goto hex_parser_exit;
-                                        }
```
microbit-carlos, line 184: This increased the flashing time of a Universal Hex by about one second with this example hex: [uhex-blocks-upy.hex.zip](https://github.com/ARMmbed/DAPLink/files/7197282/uhex-blocks-upy.hex.zip)
Mostly the time to parse content for a different target (micro:bit V1 data) goes up from ~3s to ~4s.

mathias-arm: This part is required to be able to flash the `uhex-sectors-upy.hex` file I was under the impression needed to be supported.

PR #853, opened 2021-09-20
Fix MSD flashing for kl26z (microbit)

mathias-arm:
I bisected to find out when `kl26z_microbit_if` MSD flashing stopped working (DAPLink crashes). This led me to [f1bb93e2]. The attached heavy-handed change fixes for gcc and armcc.

@gerargz, @flit, @carlosperate: other nRF51822 platforms seem to work without this patch, but I am not sure why why it only impacts kl26z.


microbit-carlos:
@gerargz do you have a bit more info about why this addition was required?

flit:
Do you know why DAPLink crashes with this code in place? What are the fault syndrome register values, and which instruction faulted? Could it be a stack overflow?

If you're getting an imprecise fault, you can add this code to init to force precise faults (at small performance penalty):

```c
    // Disable write buffer to make bus faults precise.
    SCnSCB->ACTLR = SCnSCB_ACTLR_DISDEFWBUF_Msk;
```

(Frankly, it's bad that DAPLink wasn't powering down debug after MSD programming all along.)

mathias-arm:
This issue might be relevant: https://github.com/microbit-foundation/DAPLink/issues/9

PR #855, opened 2021-09-21
Recover performance for block-aligned uhex images (fixes #847)

### LINKED ISSUE #847: micro:bit: Universal hex flashing on develop

mathias-arm:
Addresses an performance regression mentioned in #847

@microbit-carlos


microbit-carlos (approved):
Can confirm this mitigates some (or all, hard to tell) of the performance lost with the previous fix.
We are still behind flashing times of the 255 release, but this also affects normal Intel Hex files, so something else might have affected this.

Thanks Mathias! 👍

PR #863, opened 2021-10-10
Set version in target configs.

flit:
- Defines a missing current version enum for the target config version.
- Sets the version field in all target config instances.
- Renames `VENDOR_TO_FAMILY()` macro to `CREATE_FAMILY_ID()` so the purpose is more clear.

### DIFF
--- a/source/board/k20dx_bl.c
+++ b/source/board/k20dx_bl.c
@@ -42,14 +42,15 @@ static const sector_info_t sectors_info[] = {
 
 // k20dx128 target information
 target_cfg_t target_device = {
+    .version                    = kTargetConfigVersion,
     .sectors_info               = sectors_info,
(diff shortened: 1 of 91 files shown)


flit:
This is a prerequisite for the CMSIS-DAP v2.1 changes.

mathias-arm:
Looking at the CMSIS-DAP v2.1 changes, I don't see where the family configuration version is needed.

I don't think we need to rely on the target version field for the CMSIS-DAP v2.1 changes, I don't think we are ready to define an ABI that would work across versions.

mathias-arm:
> I thought about your suggestion to rename `kTargetConfigVersion` to `kTargetConfigCurrentVersion`. If we do this, then we need to do the same for `kBoardInfoVersion`. That affects all board files. Let me know if you want it.

Right now there is only one value, we can revisit naming when we have more clarity on what additional values will be in practice.

PR #864, opened 2021-10-10
details.txt: change DAPLink URL to daplink.io.

flit:


mathias-arm:
I think this should also modified for `daplink_target_url` values in the bootloaders and probably `USBD_WEBUSB_*_URL` constants.

PR #865, opened 2021-10-12
Merge MAX32620/30/60 target reset feature

ozersa:
Signed-off-by: Sadik.Ozer <Sadik.Ozer@maximintegrated.com>

### DIFF
--- a/records/board/max32620fthr.yaml
+++ b/records/board/max32620fthr.yaml
@@ -5,5 +5,3 @@ common:
         family:
             - source/family/maxim/max32620/target.c
             - source/family/maxim/target_reset_max32xxx.c
-    macros:
-        - TARGET_MXM=32620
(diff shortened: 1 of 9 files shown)


ozersa:
Thanks for feedback, 
I am trying to remove code duplication,
The macro define in .yaml file

https://github.com/ARMmbed/DAPLink/pull/865/commits/533760882c2b45d9fc4da7348cc5b2208780a90c#diff-f3a06f2092799268f0e3d364246ca4706cf8a5f60d641a3cd7cef5bcebb78ffcR9

Is this method forbidden? I would appreciate if I learn reason.
Thank you.

mathias-arm:
Having macro defines in the board .yaml file is allowed, but it should not affect families. They should be added to `records/family/all_family.yaml` which ends up being included in many projects.

ozersa:
Can it be merged as it is?

mathias-arm:
The goal of families is to build target-independent interface firmwares. One of the goal is to be able to support assembling a HIC interface firmware base with a target-specific part in post-processing. It would also a necessary to an interface firmware that would detect the target and automatically configure itself accordingly.

ozersa:
Thanks for the information, 
Related modification has been applied.

PR #882, opened 2021-10-22
Add nrf52820_hic

mathias-arm:
Add HIC based on Nordic Semiconductor nRF52820. It also supports Nordic Semiconductor nRF52833 as a drop-in replacement with dynamic pin assignment.


0Grit:
Awesome work. I can't wait for this.

PR #884, opened 2021-10-24
const_mbr: Initialize MBR at compile-time

mbrossard:
This saves `VFS_SECTOR_SIZE` (512) bytes of RAM by pre-computing all the fields that depend on the virtual disk  which is known (`VFS_DISK_SIZE`) at compile-time.


mbrossard:
Replaced by #1000

PR #904, opened 2021-12-05
SWO support for LPC4322 and other enhancements

mbrossard:
This is a refresh of #813 from @groleo with:
- Added SWO support
- Updated startup / system
- Improved GPIO
- Increase core frequency to 120 Mhz

Also thanks to @kc8apf who did another SWO effort and pointed out that the latest CMSIS UART drivers did not work.

PR #712, opened 2020-05-01
Add SWO support for MIMXRT1170-EVK

groleo:
This p-r is adding support for NXP's i.MXRT1170-EVK board
and enables SWO for it.

PR #906, opened 2021-12-06
Add support for MIMXRT1170-EVK and MIMXRT1060-EVK boards

mbrossard:
Builds on #904 PR.

MIMXRT1170-EVK work initially done by @nxpyandld and @groleo (with CMSIS-DAP 2.1 refresh added).


mbrossard:
In #712, @groleo wrote about the changes in `source/family/freescale/target_reset_mimxrt.c`:
> From what I understand, if buf is a FlexSPI Config block (the "FCFB"/0x42464346 test), it moves the flash start
> address;  but I have no idea what the test against 0xFFFFFFFF part does; I see some other platforms have a
> "cookieList"

@nxpyandld, @groleo, @flit:  Is that something that will interfere with other i.MX RT targets?

PR #910, opened 2021-12-08
tools: add exit error codes when a project name is invalid or git throws errors

microbit-carlos:
- Add error exit code when project name provided is invalid.
    -  If project_generator doesn't find the project you get an empty generator, it logs something like this, and progen_compile.py exits 0:
      `2021-12-08 18:01:30,260                 root ERROR      You specified an invalid project name.`
- Check error code from generate_version_file().


mathias-arm:
I think we will delete `mbedcli_compile.py` soon. Is it even still working? Anyways, you did not need to enhance it.

PR #911, opened 2021-12-09
DAP: fill the UART capability info only when DAP_UART=1

groleo:
OpenOCD ends up considering info[1] as capabilities, which in this
case is equal zero.

Signed-off-by: Adrian Negreanu <adrian.negreanu@nxp.com>


elfmimi:
I confirm that with this patch applied lpc11u35_if becomes usable by openocd-0.11.0 . tested on both windows and linux.

Plus , I created lpc11u35_if without bulk support by removing this line `- records/usb/usb-bulk.yaml` and then it becomes usable by openocd-0.10.0 as well. tested on both windows and linux.

PR #912, opened 2021-12-16
Microbit v2: Add HW timer for delays instead of busy "for" loops

gerargz:
Created as draft to wait for @microbit-carlos testing with armcc.

### DIFF
--- a/source/board/microbitv2/kl27z/pwr_mon.c
+++ b/source/board/microbitv2/kl27z/pwr_mon.c
@@ -25,6 +25,7 @@
 #include "fsl_pmc.h"
 #include "fsl_port.h"
 #include "fsl_gpio.h"
+#include "fsl_tpm.h"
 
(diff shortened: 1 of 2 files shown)


microbit-carlos:
Measurements of the two delays 3ms (used in several places) and 100 us. 
|              | ARMCC busy loop  | GCC  timer |
|--------|--------|-------|
| 3ms delay | image 3.023ms |  image 3.025 ms |
 | 100us |  image 100.9 us |  image 102.3us |


I must admit I'm a bit surprised the busy loop is so accurate, but I guess it makes sense if each loop iteration is exactly 3 cycles.
The 100us delay with the timer is a little bit off, probably the overhead of the function call. @gerargz If we reduce that to 98 we get 100.3 us:
image

microbit-carlos:
Okay, so I've reduced the 100us delay and this PR should be ready for merge.

Issue #915, opened 2021-12-28
DAPv1 bundled with DAPv2 has compatibility issues


elfmimi:
HID interface of DAPLink may not functioning as expected when bulk interface is also present.
It seems there are more than one pattern of behaviors.
I've analyzed these two out of many.
- firmware-0257-798b9e73/gcc/0257_lpc11u35__0x0000.hex
  HID interface does not have any endpoint, because of endpoint shortage.\
  Host must use SET_REPORT and GET_REPORT to interact with this.\
  Not many software is aware of this kind of configuration.\
  Even Linux won't bind hid driver for this device.

- ~~firmware-0257-798b9e73/gcc/0257_lpc55s69__0x10000.hex
  HID interface has both IN and OUT endpoints.\
  Despite having OUT endpoint it will get stuck if try to use it.\
  Host must use SET_REPORT to workaround the problem.~~
  Solved with https://github.com/ARMmbed/DAPLink/pull/917

mbrossard:
LPC55S69 HIC fix has been merged. `lpc11u35_if` is the only project using the `lpc11u35` HIC that has bulk interface. There might be 3 simple solutions:
- Remove bulk interface from `lpc11u35_if`.
- Remove HID interface from `lpc11u35_if`.
- Remove bulk interface from `lpc11u35_if` and add `lpc11u35_bulk_if` with bulk (instead of HID) interface.

mbrossard:
I tried creating a project with bulk interface but without HID interface and it did not work, so as a first step #920 changes `lpc11u35_if` to be HID-only, like all other `lpc11u35`-based projects.

elfmimi:
I don't yet know why, but this must be the cause bulk interface without HID interface is failing.
https://github.com/ARMmbed/DAPLink/blob/8dff8e01027fdb23c49734099a61ee1077af0797/source/hic_hal/nxp/lpc11u35/DAP_config.h#L97-L101
Try changing it to 64 , then it should work fine.

mbrossard:
Looking at the USB configuration it seems all interefaces are limited to 64 bytes:
```$ grep WMAXPACKETSIZE source/hic_hal/nxp/lpc11u35/usb_config.c | grep -v _HS_
#define USBD_HID_WMAXPACKETSIZE     64
#define USBD_MSC_WMAXPACKETSIZE     64
#define USBD_ADC_WMAXPACKETSIZE     64
[...]

PR #917, opened 2021-12-29
Fix HID interface of LPC55xx HIC

elfmimi:
Fixes DAPv1 compatibility issue of lpc55s69_if mentioned in https://github.com/ARMmbed/DAPLink/issues/915 .
This also makes it compatible with now old openocd-0.10.0 .


flit (approved):

flit:
Really good fix, @elfmimi! 💯 

It looks like the same change should be applied to the LPC11U35 USB driver, but only for completeness. In that case the bug breaks isochronous EPs and interrupt EPs in rate feedback mode, neither of which are used by DAPLink. But it wouldn't be good to leave a known bug residing in the code.

PR #921, opened 2022-01-04
microbit: Set CDC ACM default baud rate for micro:bit v1 and v2.0.

microbit-carlos:
And remove BOARD_EXTRA_BUFFER vestiges that were not used.


microbit-carlos:
Rebased.

PR #924, opened 2022-01-06
lpc4322_hic: make use of OS_CLOCK defined in yaml

groleo:


mathias-arm:
I have also wondered if we should not replace `SystemCoreClock` with `OS_CLOCK` there too: https://github.com/ARMmbed/DAPLink/blob/da0afc79ab343c7c24cdb91784aa81818f180173/source/hic_hal/nxp/lpc4322/DAP_config.h#L55
cc: @flit

groleo:
> I have also wondered if we should not replace `SystemCoreClock` with `OS_CLOCK` there too:

My guess is that when `SystemCoreClock` is used, its value is re-read it in case it was modified. (again, just a guess)

mathias-arm:
> My guess is that when `SystemCoreClock` is used, its value is re-read it in case it was modified. (again, just a guess)

That's my guess too, but ideally we should be able to decide a frequency. The difference between a variable and a constant might be important in this case, because this is used by CMSIS-DAP for timing SWD/JTAG bit-banging.

groleo:
Moved to #925

PR #937, opened 2022-02-03
microbitv2: Reduce target SWD clock to 6 MHz

gerargz:


microbit-carlos:
@gerargz @mathias-arm do you remember what issue was resolved by reducing the SWD speed on V2.00 KL27?

PR #946, opened 2022-02-25
fix flash algo generation script

zjli-2019:
1. generate `#define` instead of `const uint32_t`;
1. add the missing field `algo_flags`.

### DIFF
--- a/tools/generate_flash_algo.py
+++ b/tools/generate_flash_algo.py
@@ -120,7 +120,9 @@ static const program_target_t flash = {
     // address of prog_blob
     {{name}}_flash_prog_blob,
     // ram_to_flash_bytes_to_be_written
-    {{'0x%08x' % algo.page_size}}
+    {{'0x%08x' % algo.page_size}},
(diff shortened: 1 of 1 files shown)


mbrossard:
It's not clear to me what you are trying to fix. I don't see any targets using `FLASH_START` / `FLASH_SIZE`. `flash_start` and `flash_size` might be problematic, but macros are not the solutions. I would rather add the region size to `sectors_info` but this is a larger project.

Making the flags explicit would be cleaner.

zjli-2019:
1. A possible use case of `flash_start` and `flash_size`:

    ```c
    target_cfg_t target_device = {
        .sectors_info                   = sectors_info,
        .sector_info_length             = (sizeof(sectors_info))/(sizeof(sector_info_t)),
        .flash_regions[0].start         = flash_start,
        .flash_regions[0].end           = flash_start + flash_size,
        //...
    ```
1.  `flags` is supposed to be edited manually later, which is not the best solution, but slightly better than none.

PR #948, opened 2022-02-25
fix compile without DRAG_N_DROP_SUPPORT

zjli-2019:


mbrossard:
Without `DRAG_N_DROP_SUPPORT` this file will not be compiled, so this change on its own does not to make sense to me.

zjli-2019:
About one year ago, with this patch, this file can be compiled without `DRAG_N_DROP_SUPPORT`. 😞

Issue #745, opened 2020-12-03
possible issue for Drag and Drop feature on Mac OS


oxanagig:
The drag and drop code logic seems to rely on that the bin/hex file is transferred to contiguous sectors.

static void file_data_handler(uint32_t sector, const uint8_t *buf, uint32_t num_of_sectors)
{
    stream_type_t stream;
    uint32_t size;

    // this is the key for starting a file write - we dont care what file types are sent
    //  just look for something unique (NVIC table, hex, srec, etc) until root dir is updated
    if (!file_transfer_state.stream_started) {
        // look for file types we can program
        stream = stream_start_identify((uint8_t *)buf, VFS_SECTOR_SIZE * num_of_sectors);

        if (STREAM_TYPE_NONE != stream) {
            transfer_stream_open(stream, sector);
        }
    }

    if (file_transfer_state.stream_started) {
        // Ignore sectors coming before this file
        if (sector < file_transfer_state.start_sector) {
            return;
        }

        // sectors must be in order
        if (**sector != file_transfer_state.file_next_sector**) {
            vfs_mngr_printf("vfs_manager file_data_handler sector=%i\r\n", sector);

            if (sector < file_transfer_state.file_next_sector) {
                vfs_mngr_printf("    sector out of order! lowest ooo = %i\r\n",
                                file_transfer_state.last_ooo_sector);

                if (VFS_INVALID_SECTOR == file_transfer_state.last_ooo_sector) {
                    file_transfer_state.last_ooo_sector = sector;
                }

                file_transfer_state.last_ooo_sector =
                    MIN(file_transfer_state.last_ooo_sector, sector);
            } else {
                vfs_mngr_printf("    sector not part of file transfer\r\n");
            }

            vfs_mngr_printf("    discarding data - size transferred=0x%x, data=%x,%x,%x,%x,...\r\n",
                            file_transfer_state.size_transferred, buf[0], buf[1], buf[2], buf[3]);
            return;
        }

        // This sector could be part of the file so record it
        size = VFS_SECTOR_SIZE * num_of_sectors;
        file_transfer_state.size_transferred += size;
        file_transfer_state.file_next_sector = sector + num_of_sectors;

        // If stream processing is done then discard the data
        if (file_transfer_state.stream_finished) {
            vfs_mngr_printf("vfs_manager file_data_handler\r\n    sector=%i, size=%i\r\n", sector, size);
[...]

PR #971, opened 2022-07-22
Add support for UF2 format

xingrz:
This PR implemented the drag-n-drop support for [UF2](https://github.com/microsoft/uf2), which is mentioned in https://github.com/ARMmbed/DAPLink/issues/745#issuecomment-738230690.

### DIFF
--- a/source/daplink/drag-n-drop/file_stream.c
+++ b/source/daplink/drag-n-drop/file_stream.c
@@ -24,6 +24,7 @@
 #include "file_stream.h"
 #include "util.h"
 #include "intelhex.h"
+#include "uf2.h"
 #include "flash_decoder.h"
 #include "error.h"
 #include "cmsis_os2.h"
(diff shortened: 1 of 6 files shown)


mbrossard:
This is a really exciting addition.

What tools for generating UF2 files have you tested this patch with?

xingrz:
> What tools for generating UF2 files have you tested this patch with?

Tested with UF2 built by Zephyr, which internally uses [uf2conf.py](https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/build/uf2conv.py)

mbrossard:
I started adding UF2 target binaries for the Zephyr DAPLink Validation binaries (https://github.com/mbrossard/zdv/commit/1d5689acc798534f472246191eaaf918f411b13a).

Zephyr's build system insists on having a Family ID. It brings some interesting questions how DAPLink should handle Family ID.

source/daplink/drag-n-drop/file_stream.c
```diff
@@ -364,3 +373,50 @@ static error_t close_hex(void *state)
...
+}
+
+static error_t write_uf2(void *state, const uint8_t *data, uint32_t size)
+{
+    error_t status;
+    uint32_t start_addr;
+    const flash_intf_t *flash_intf;
+    const UF2_Block *block;
+
+    if (1 != validate_uf2block(data, size)) {
```
mathias-arm, line 398: This code assumes the calls to `write_uf2()` are one sector (512 bytes) long. Looking at `file_data_handler()` in `source/daplink/drag-n-drop/vfs_manager.c`, it looks like it could be multiple sectors (though I am not sure this can actually happen).
There is also the CMSIS-DAP / WebUSB commands (see `source/daplink/cmsis-dap/DAP_vendor.c`) for flashing files, because it works over the HID endpoint will (once framing has been removed) write about 40 bytes at a time.
We might need a buffer (the `shared_state_t` variant from hex files already has a 256 bytes buffer).

PR #1008, opened 2023-02-04
CMSIS-DAPv2: blink LED on bulk interface activity

mbrossard:
Fix for #714

### DIFF
--- a/source/usb/bulk/usbd_bulk.c
+++ b/source/usb/bulk/usbd_bulk.c
@@ -25,6 +25,8 @@
 #include "usb_for_lib.h"
 #include "util.h"
 #include "DAP_queue.h"
+#include "daplink.h"
+#include DAPLINK_MAIN_HEADER
 
 static U8 *ptrDataIn;
 static U16 DataInReceLen;
@@ -76,6 +78,7 @@ void USBD_BULK_EP_BULKOUT_Event(U32 event)
     if ((DataInReceLen >= USBD_Bulk_BulkBufSize) ||
             (bytes_rece    <  usbd_bulk_maxpacketsize[USBD_HighSpeed])) {
         if (DAP_queue_execute_buf(&DAP_Cmd_queue, USBD_Bulk_BulkOutBuf, DataInReceLen, &rbuf)) {
+            main_blink_hid_led(MAIN_LED_FLASH);
             //Trigger the BULKIn for the reply
             if (USB_ResponseIdle) {
                 USBD_BULK_EP_BULKIN_Event(0);


microbit-carlos:
As it turns out, the micro:bit online editors (MakeCode & Python) both constantly pull serial data when connected via WebUSB. This is done to offer some useful features like detecting errors printed to serial and highlighting them in the editor.

With this fix, unfortunately, the editors are constantly blinking the DAPLink activity LED as soon as the micro:bit is WebUSB connected, and therefore we cannot tell the difference between the micro:bit seating idle vs flashing (or the user interacting with the target serial), which is technically worse for our users, as seeing the flashing LED activity is quite useful and the behaviour we've described in a lot of the micro:bit documentation.

One option we are considering is seeing if there is a way to check the bulk activity traffic, and if possible maybe filter out activity coming from serial data (of course, that would have to be behind a build flag or hook).
@mathias-arm Until we have a chance to investigate a workaround like this, could we have this fix behind a build flag?

mbrossard:
Do you think this could work: https://github.com/mbrossard/DAPLink/tree/pr/uart-dap-led?

microbit-carlos:
Thanks @mbrossard, unfortunately that didn't work. I think `DAP_ProcessVendorCommand()` likely executes inside `DAP_queue_execute_buf()`, so `main_blink_hid_led(MAIN_LED_FLASH)` would overwrite the state set in that patch: 
https://github.com/ARMmbed/DAPLink/blob/5e0412379644cdfde8662a18b6cd018011257ed5/source/usb/bulk/usbd_bulk.c#L80-L81

Doing this as quick hack did work (vendor IDs 3 and 4 are serial read/write), but not sure how to best integrate it in DAPLink.
https://github.com/ARMmbed/DAPLink/commit/236e1c4310e3ebaa4d12a8e8b8105eb9eb8cf760

microbit-carlos:
The call stack goes:
- `DAP_queue_execute_buf()` ->
    - `DAP_ExecuteCommand()` ->
        - Here It can execute `DAP_ProcessCommand()` one or multiple times ->
            - Here the DAP command IDs are checked for the first time and processed, and it might call `DAP_ProcessVendorCommand()`

I guess we could probably add `main_blink_hid_led(MAIN_LED_FLASH)` at the top of `DAP_ProcessCommand()` and `main_blink_hid_led(MAIN_LED_DEF)` when `DAP_ProcessVendorCommand()` processes the serial read/write DAP commands?

mathias-arm:
I think we could move the `main_blink_hid_led()` calls in `source/usb/bulk/usbd_bulk.c` and `source/usb/hid/usbd_user_hid.c` to be in `DAP_queue_execute_buf()` and have the `usbd_hid_no_activity()` apply to both HID and Bulk.

I prefer modifying `DAP_queue_execute_buf()` because it's a DAPLink function, when `DAP_ExecuteCommand()` is from CMSIS-DAP.

microbit-carlos:
We could use a similar hook to `usbd_hid_no_activity` in `source/usb/bulk/usbd_bulk.c`:
https://github.com/microbit-foundation/DAPLink/commit/ef02cf2f4d2bfbb17615178c7726bdebc2eba52c

Ideally we'd use the same hook, but if the function declaration needs to go into a header file I'm not sure what's the best place to put it, any suggestions?

mathias-arm:
I agree there should only be one hook. The "HID" LED should rather be considered the "DAP" LED. So I would be in favor of renaming `usbd_hid_no_activity()` to `usbd_dap_no_activity()` for instance. Both `source/usb/bulk/usbd_bulk.c` and `source/usb/hid/usbd_user_hid.c` import `DAPLINK_MAIN_HEADER` (`main_interface.h`) for `main_blink_hid_led()`, so that would be my suggestion.

I think the default hook could ignore the DAP commands that call `main_blink_cdc_led()`.

microbit-carlos:
The easiest way to use a single hook might be to move it to `DAP_queue_execute_buf()` as you suggested before.

Which approach would you prefer then:
- Single hook called in `usbd_bulk.c` and `usbd_user_hid.c` that might require the weak function to be declared in a header file
    - Similar to https://github.com/microbit-foundation/DAPLink/commit/86ab3992550fa11b05fa8dff77757c10ad802cd7
    - Which header file would the declaration go? `rl_usb.h` is likely an RTX header that shouldn't be touched, so the alternative might be to create a new one?
- Single hook moved to `DAP_queue.c`
    - As shown in https://github.com/microbit-foundation/DAPLink/commit/c718b870d499ba79dbe94da54c294c14c4cb252d

Or something else?

microbit-carlos:
Okay, I've created both PRs, feel free to merge whichever you prefer if it's ready to go (comments with improvements, maybe with the naming, are welcomed).

- https://github.com/ARMmbed/DAPLink/pull/1017
- https://github.com/ARMmbed/DAPLink/pull/1018

PR #1016, opened 2023-03-20
nrf52: Add new flash algo for nRF52833 that improves flashing time.

microbit-carlos:
More info about the flash algo improvements in:
- https://github.com/pyocd/FlashAlgo/pull/68

As the improvement was achieved via a feature not present in all members of the nRF52 family, the original nRF52832 flash algo stays and this version for nRF52833 has been added along side it.

Could potentially be used for nRF52840 targets as well.


mathias-arm (approved):

mathias-arm:
It would be interesting to enable this for nRF52840 if it works and gives better performances.

PR #1017, opened 2023-03-22
Update hook to skip blinking on WebUSB HID serial traffic (option 1).

microbit-carlos:
Rename the usbd_hid_no_activity hook to main_dap_no_activity as it has been moved to the main_interface.h header file.

Add check to usbd_bulk.c to also skip blinks with bulk traffic, if applicable to the project.

-----

This is one option out of two, so between https://github.com/ARMmbed/DAPLink/pull/1018 and this only one PR should be merged.

### DIFF
--- a/source/board/artemis_dk.c
+++ b/source/board/artemis_dk.c
@@ -23,6 +23,7 @@
 #include "DAP.h"
 #include "target_family.h"
 #include "target_board.h"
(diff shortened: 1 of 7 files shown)


mathias-arm:
Let's go with https://github.com/ARMmbed/DAPLink/pull/1018

PR #1018, opened 2023-03-22
Update hook to skip blinking on WebUSB HID serial traffic

microbit-carlos:
Move usbd_hid_no_activity() hook from usbd_user_hid.c to DAP_queue.c and rename it to DAP_no_activity().

In this new location it can be used to skip HID LED blinking in traffic from USB bulk as well.

-----

This is one option out of two, so between https://github.com/ARMmbed/DAPLink/pull/1017 and this only one PR should be merged.

PR #1019, opened 2023-03-22
Update hook to skip blinking on WebUSB HID serial traffic.

microbit-carlos:
A continuation from this PR, opened from my account fork so that it can be edited by maintainers:
- https://github.com/ARMmbed/DAPLink/pull/1018

Implemented the suggestion from https://github.com/ARMmbed/DAPLink/pull/1018#issuecomment-1480288695 and the final result is:

Moves `usbd_hid_no_activity()` hook from `usbd_user_hid.c` to `DAP_queue.c` and renames it to `DAP_activity_blink()`.

In this new location it can be used to skip HID LED blinking in traffic from USB bulk as well.

Adds the default behaviour to the weak function to skip HID blink of DAP serial & MSD activity.

Removes customised hooks in the artemis & micro:bit projects as they are no longer needed.

The logic in the hook has been reverted from a "negative check" to a "positive check" (before `usbd_hid_no_activity()` returned true when there was no activity, so then we checked for "not [no activity]") as this way it's easier to reason.

### DIFF
--- a/source/board/artemis_dk.c
+++ b/source/board/artemis_dk.c
@@ -52,15 +52,6 @@ static void prerun_board_config(void)
     set_board_id(board_version);
 }
 
-// USB HID override function return 1 if the activity is trivial or response is null
-uint8_t usbd_hid_no_activity(uint8_t *buf)
-{
-    if (buf[0] == ID_DAP_Vendor3 && buf[1] == 0)
-        return 1;
-    else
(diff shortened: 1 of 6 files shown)


microbit-carlos:
@mathias-arm you should be able to edit this branch before merging, so feel free to do any changes you think are appropriate 👍

PR #1022, opened 2023-04-01
Allow use of MSC/CDC USB endpoints for bulk on sam3u2

9names:
The sam3u2 USB peripheral does not have enough endpoints after CDC and MSC are enabled to enable the BULK endpoints. Bulk endpoints are required for CMSIS-DAP v2 support - the existing projects for this HIC use HID+MSC+CDC

This PR makes it possible for the sam3u2 to enable BULK if either CDC or MSC are disabled.
In order to demonstrate this setup, the sam3u2c_bulk_test_if project has been added which uses HID+CDC+BULK, 

I have tested this configuration with a nanoDAP-hs, probe-rs-cli and few computers I had at my disposal:
raspberry pi 4 -  hid: 90KB/s, bulk: 152KB/s
AMD 7700x, ubuntu 22.10 -  hid: 17KB/s, bulk: 280KB/s
intel u5200, windows 10 - hid: 100KB/s, bulk: 160KB/s

Not sure why my AMD box is so slow with HID, but the speed-up is very significant in this case.

### DIFF
--- a/projects.yaml
+++ b/projects.yaml
@@ -37,6 +37,22 @@ module:
         - records/daplink/interface.yaml
         - records/daplink/target_family.yaml
         - records/daplink/target_board.yaml
+    # this is exactly the same as if without usb-msc
+    # intended use is HICs that do not have enough USB endpoints to have MSC and BULK at the same time
+    no_msc_if: &module_no_msc_if
+        - *module_tools
+        - records/usb/usb-core.yaml
+        - records/usb/usb-cdc.yaml
+        - records/usb/usb-webusb.yaml
+        - records/usb/usb-winusb.yaml
+        - records/daplink/cmsis-dap.yaml
+        - records/daplink/drag-n-drop.yaml
+        - records/daplink/usb2uart.yaml
+        - records/daplink/settings.yaml
+        - records/daplink/settings_rom.yaml
+        - records/daplink/interface.yaml
+        - records/daplink/target_family.yaml
+        - records/daplink/target_board.yaml
     # HICs
     hic_k20dx: &module_hic_k20dx
         - records/rtos/rtos-cm3.yaml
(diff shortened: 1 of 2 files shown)


9names:
Would it make more sense to move usb-msc out of `bl` and `if`, and have it added by default to each `HIC` where it currently exists (except for sam3u2)?
The `HIC`s already add `usb-bulk` and `usb-hid` themselves, so it would seem reasonable to include `usb-msc` there as well.

mbrossard:
The additive nature of DAPLink's build configuration is problematic for interface that have limited amount of endpoints. I did also consider your idea of splitting further the configuration of `bl` / `if` projects, but I got worried it would grow the verbosity of the `projects.yaml`. Instead I am leaning towards adding the option to disable functionality from the default interface configuration. I ported some of the changes I had started to make onto  the [feature/flexible-usb-configuration](https://github.com/mbrossard/DAPLink/tree/feature/flexible-usb-configuration) branch and modified `sam3u2c_if` to disable `MSC` endpoint and replace `HID` with `BULK`. I am not 100% sure about modifying `sam3u2c_if` rather than adding a new project, but this is what I am leaning towards.

I would have usually preferred to go with your change to `usb_config.c` (which is nicely compact). But in this case where things are a little bit complicated (I don't understand how things work with `USBD_HID_EP_INTOUT` being set to `0`), I would rather be very explicit on the endpoint assignments.

9names:
Your version looks better to me :)

I opted out of modifying existing projects or removing HID because I was worried existing users might expect everything to stay the same, but I assume you know your users better than I do.
Turning off HID when BULK is enabled is something I would have done if it were as simple as it is for your branch.

If this were a pure cmake project, I would have made these things selectable at configure time and overridable at build time, but I don't know pygen well enough to know what is possible and what is considered idiomatic usage

mathias-arm:
Replaced by #1041

PR #1031, opened 2023-06-15
Actually pull up nRESET on STM32F103

### LINKED ISSUE #824: STM32F103 HIC nRESET pin pullup issue
I tested DAPLink on HIC with STM32F103CB without problems. Using HIC with clone GD32F103CB causes problem - it starts in bootloader update mode. I investigated this issue and I found the reason.
At startup bootloader checks nRESET pin to determine to go in update mode if this pin is shorted to ground. In theory nRESET pin is configured to output OD state with pullup. I checked datasheet and found that unfortunatelly pins in STM32F103 (and also in clones) have hardware disabled internal pullup or pulldown when they are in in output state. This mode can't be enabled like in other microcontrollers series like STM32F0.
nRESET pin at startup (after configuration) acts as floating pin with undetermined state. Statistically on STM32F103 it is read as Hi state so it seems that programmer works, on GD32F103 it almost always is read as Lo state so programmer goes to update mode.
The best solution is to rewrite code and define nRESET pin output states as follows:
Lo - output OD mode with Lo state
Hi - input mode with pullup
The simplest solution I've found is to change configuration in "hic_hal\stm32\stm32f103xb\gpio.c" file from
```
    // reset button configured as gpio open drain output with a pullup
    HAL_GPIO_WritePin(nRESET_PIN_PORT, nRESET_PIN, GPIO_PIN_SET);
    GPIO_InitStructure.Pin = nRESET_PIN;
    GPIO_InitStructure.Mode = GPIO_MODE_OUTPUT_OD;
    GPIO_InitStructure.Pull = GPIO_PULLUP;
    HAL_GPIO_Init(nRESET_PIN_PORT, &GPIO_InitStructure);
```
to
```
    // reset button configured as gpio input with a pullup (STM32F103 doesn't have pullup in output OD mode)
    HAL_GPIO_WritePin(nRESET_PIN_PORT, nRESET_PIN, GPIO_PIN_SET);
    GPIO_InitStructure.Pin = nRESET_PIN;
    GPIO_InitStructure.Mode = GPIO_MODE_INPUT;
    GPIO_InitStructure.Pull = GPIO_PULLUP;
    HAL_GPIO_Init(nRESET_PIN_PORT, &GPIO_InitStructure);
```
This pin is reconfigured to output OD mode in functions DAP_SETUP() and PORT_SWD_SETUP() so this solution is safe and allows to check nRESET pin real state at startup. For me it works.

ducky64:
Despite the comments in the STM32F103 HIC, the nRESET pin was never pulled up since the STM32F103 can't do simultaneous open drain and pull up. Previously, it relies on an external pullup resistor to not enter bootloader mode, which isn't present on some standalone debug probes.

This changes the code to initialize nRESET in input mode with pull up when nRESET is high, and configures it in open-drain mode when nRESET is asserted low.

Testing notes:
- Previously, running DAPLink on a STM32F103-based standalone programmer without a nRESET pullup would cause it to always enter bootloader mode. Now, that only happens when nRESET is grounded.
- STLink V2 clone note: these devices have a different nRESET pinning, on PB5 (through a resistor) and PB6 (direct on the nRESET pin). With this PR, they will always boot into the interface, but they can't be forced back into bootloader mode by grounding the nRESET pin.
  - Things behave as expected if the pinning is corrected (and the default CONNECTED_LED is moved elsewhere from the reset pin on these boards).

Resolves #824 (wasn't clear why that never made it into a PR).

### DIFF
--- a/source/hic_hal/stm32/stm32f103xb/DAP_config.h
+++ b/source/hic_hal/stm32/stm32f103xb/DAP_config.h
@@ -260,7 +260,7 @@ __STATIC_INLINE void PORT_SWD_SETUP(void)
 
     pin_in_init(SWDIO_IN_PIN_PORT, SWDIO_IN_PIN_Bit, 1);
     // Set RESET HIGH
-    pin_out_od_init(nRESET_PIN_PORT, nRESET_PIN_Bit);//TODO - fix reset logic
+    pin_in_init(nRESET_PIN_PORT, nRESET_PIN_Bit, 1);  // input with pullup
     nRESET_PIN_PORT->BSRR = nRESET_PIN;
 }
 
@@ -436,10 +436,13 @@ __STATIC_FORCEINLINE uint32_t PIN_nRESET_IN(void)
 
 __STATIC_FORCEINLINE void     PIN_nRESET_OUT(uint32_t bit)
 {
-    if (bit & 1)
+    if (bit & 1) {  // set to input pullup mode, since open-drain doesn't support pullup
+        pin_in_init(nRESET_PIN_PORT, nRESET_PIN_Bit, 1);
         nRESET_PIN_PORT->BSRR = nRESET_PIN;
-    else
+    } else {  // set to open-drain mode
+        pin_out_od_init(nRESET_PIN_PORT, nRESET_PIN_Bit);
         nRESET_PIN_PORT->BRR = nRESET_PIN;
+    }
 }
 
 //**************************************************************************************************
@@ -536,7 +539,7 @@ __STATIC_INLINE void DAP_SETUP(void)
 
     pin_in_init(SWDIO_IN_PIN_PORT, SWDIO_IN_PIN_Bit, 1);
 
-    pin_out_od_init(nRESET_PIN_PORT, nRESET_PIN_Bit);
+    pin_in_init(nRESET_PIN_PORT, nRESET_PIN_Bit, 1);  // input with pullup
     nRESET_PIN_PORT->BSRR = nRESET_PIN;
 
     pin_out_init(CONNECTED_LED_PORT, CONNECTED_LED_PIN_Bit);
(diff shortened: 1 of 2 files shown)


llinjupt:
@ducky64 Thanks for your work. I ran across this issue as well, and a more simple solution to me is to update the gpio_init() in gpio.c:

```
#ifdef DAPLINK_BL  // Bootloader needs to read nRESET_PIN to determine if stay in bootloader
    GPIO_InitStructure.Pin = nRESET_PIN;
    GPIO_InitStructure.Mode = GPIO_MODE_INPUT;
    GPIO_InitStructure.Pull = GPIO_PULLUP;
    HAL_GPIO_Init(nRESET_PIN_PORT, &GPIO_InitStructure);
#else
    // reset button configured as gpio open drain
    HAL_GPIO_WritePin(nRESET_PIN_PORT, nRESET_PIN, GPIO_PIN_SET);
    GPIO_InitStructure.Pin = nRESET_PIN;
    GPIO_InitStructure.Mode = GPIO_MODE_OUTPUT_OD; 
    GPIO_InitStructure.Pull = GPIO_PULLUP;
    HAL_GPIO_Init(nRESET_PIN_PORT, &GPIO_InitStructure);
#endif
```
Normally, a RESET_PIN should be OD for the electrical level is decided by the MCU itself.  It got this issue because DAPLink shares the same gpio.c between bootloader and IF and luckly there are two macros DAPLINK_BL and DAPLINK_IF to identify which is which.  Anyway it's only for your reference.

mbrossard:
I think the solution from @llinjupt makes more sense. @ducky64: can you test his solution and update the PR accordingly?

ducky64:
Done and tested working on a board without a dedicated reset pullup. When powered with reset floating, it boots into DAP mode. When powered with no firmware or reset pulled low, it boots into maintenance mode.

PR #1041, opened 2023-08-04
Flexible USB Configuration and optimize `sam3u2c_if`

mbrossard:
The current build configuration makes it difficult to handle the interfaces that do not have enough endpoints or when the decision to include HID and/or BULK is done on a per-project basis. 

This PR adds the possibility to disable endpoints and some functionality through macro definitions. An example of the feature is that moving forward `sam3u2c_if` will use BULK endpoints. This requires disabling HID endpoints, which thanks to this PR is done without affecting other projects using `module_hic_sam3u2c`.


mbrossard:
This PR should supersede #1022.

PR #1043, opened 2023-08-05
Support for newer Arm GNU Toolchain versions

### LINKED ISSUE #1040: Support for newer versions of GNU Arm Embedded Toolchain
There are two link-time issues with newer versions of `arm-none-eabi-gcc`

The first one is related to some missing symbols:
```
[...]/libc_nano.a(libc_a-closer.o): in function `_close_r':
[...]/closer.c:47: warning: _close is not implemented and will always fail
[...]/libc_nano.a(libc_a-lseekr.o): in function `_lseek_r':
[...]/lseekr.c:49: warning: _lseek is not implemented and will always fail
[...]/libc_nano.a(libc_a-readr.o): in function `_read_r':
[...]/readr.c:49: warning: _read is not implemented and will always fail
[...]/libc_nano.a(libc_a-writer.o): in function `_write_r':
[...]/writer.c:49: warning: _write is not implemented and will always fail
```

A work-around for this issue is to add the following lines in a .c file (e.g. source/daplink/sdk_stub.c):
```
void _close_r(void) {}
void _lseek_r(void) {}
void _read_r(void) {}
void _write_r(void) {}
```

The other issue relates to the permission of the RAM segment (needed in some cases for RAM functions when writing to flash):
```
warning: build/xxxxx.elf has a LOAD segment with RWX permissions
```

The work-around is add `-Wl,--no-warn-rwx-segment` to `ld_flags` in `records/tools/gcc_arm.yaml`. This is not a solution we could just merge because the flag is not supported by older versions of `arm-none-eabi-gcc`. Another alternative would be to remove `-Werror`, but that seems even worse.

mbrossard:
- Fixes #1040
- Updates CI to use Arm GNU Toolchain version 13.2.Rel1 (builds on #1042)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -8,10 +8,10 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
-        gcc: ['13.2.Rel1']
+        gcc: ['14.2.Rel1']
 
     steps:
     - name: Install Arm GNU Toolchain (arm-none-eabi-gcc)
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        gcc: ['12.3.Rel1']
+        gcc: ['13.2.Rel1']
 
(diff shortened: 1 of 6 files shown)


imi415 (approved):
Confirmed work on 13.2 Rel1 as well as 14.1.0 shipped by Arch Linux.

Issue #1089, opened 2025-01-28
Interrupting a USB connection before device sends DAP response causes queue missalignment


microbit-carlos:
Tested with the latest code in the main branch and replicable with 0255 forward (not tested with older versions).

During a WebUSB connection, if the a transfer is interrupted before the device can send a DAP response (for example, by having a battery connected and yanking the USB cable), the next DAPJs connections will fail, as the DAP queue incorrectly responds to the next command with the unsent data from the previous response. This leaves to the DAP queue to always respond with stale data, being "one off".

To replicate with a micro:bit V2.2 running 0256+:
- Connect a battery pack
- Connect the USB cable
- Open MakeCode v7 https://makecode.microbit.org/v7#editor
    - The MakeCode editor constantly checks in the background if there is any UART data to read
- Open the developer console
- Connect via WebUSB and flash
- Try multiple times: Disconnect and reconnect the USB cable
    - After inserting the USB cable, MakeCode should try to automatically reconnect WebUSB. Wait until the micro:bit logo on the "Download" button stops moving (it animates during WebUSB connection)
    - Every time the cable is disconnected at this point it should print an error in the console, this indicates a USB transfer was interrupted before the response was received.
- Eventually, after a few reconnections the the WebUSB connection will throw an error on the console and won't connect.

We were familiar with this issue for a while, and was reported here:
- https://github.com/microbit-foundation/DAPLink/issues/17
Opening this issue upstream to open a PR and link it.

PR #1090, opened 2025-01-28
Fix abrupt USB disconnection leaving stale data in DAP queue buffers.

microbit-carlos:
Clearing the buffers in the `USB_DISCONNECTING` state, which then moves to the `USB_DISCONNECTED` state.

The `USB_DISCONNECTING` state was never used, in the current codebase nor the initial repo commit (well, second commit, first release): bdacee77a463259046bd70fd9551651f8d7a95a0 ([this .patch URL is easier to search in-browser](https://github.com/ARMmbed/DAPLink/commit/bdacee77a463259046bd70fd9551651f8d7a95a0.patch))

It looks like this state might have been initially designed as a way to trigger a USB disconnection from the interface chip, as it manually set `usbd_connect(0)` to disconnect.
As it was never used it's now the pre-disconnected state to run code only once on state change:
connected->disconnecting->disconnected

Fixes https://github.com/ARMmbed/DAPLink/issues/1089


microbit-carlos:
I've found an issue in this PR with the KL27 of the micro:bit V2 + battery pack, where sometimes inserting the USB cable doesn't start the USB connection with the PC (might be related to the voltage monitoring interrupt, which is meant to trigger this).

PR #1092, opened 2025-02-03
Process out-of-order vfs sectors for Universal Hex in block format

microbit-carlos:


mathias-arm:
I merged #1097, so you can rebase.

Could you have a look at https://github.com/mbrossard/DAPLink/tree/feature/vfs-tests? It's a way to unit test a part of the flashing logic. I think you should cherry-pick the `vfs-tests` commits and document how to download or build the different Universal Hex files.

source/board/microbitv2/microbitv2.c
```diff
@@ -446,13 +446,16 @@ static uint32_t read_file_data_txt(uint32_t sector_offset, uint8_t *data, uint32
     return VFS_SECTOR_SIZE;
 }
 
-uint8_t board_detect_incompatible_image(const uint8_t *data, uint32_t size)
+uint8_t board_detect_incompatible_image(uint32_t addr, const uint8_t *data, uint32_t size)
 {
-    uint8_t result = 0;
+    // We can only check image compatibility on the first block of flash data
+    if (addr != 0) {
```
mathias-arm, line 452: The first block of flash data is not always at address 0x0.

microbit-carlos: Right, this didn't meant the first block received that contains flash data, but that the check should only be performed on the block that contains the first 512 bytes of flash data, which might arrive at any point.
I can update the comment to make this more clear.

microbit-carlos:
Thanks @mathias-arm!
I've removed the PR https://github.com/ARMmbed/DAPLink/pull/1097 commit, rebased, and force pushed. I'll look into the `vfs-tests` once I find some time this week.

source/board/microbitv2/microbitv2.c
mathias-arm: I think `addr` is not the offset in the image, but the address at which the block is intended to be flashed.

microbit-carlos: Right, I'll need to double check this, but my assumption here is that as this is being called within `flash_decoder_write(addr, data, size)`, that would be the address in the target flash where the data block will be written.
So, this function is meant to let pass through all other data blocks as being compatible, and only perform the check on the raw data going to flash address 0x0. At that point it checks the vector table to see if it's an M4 or M0+ image and if it fails the check it makes DAPLink stop flashing and throw an error.

This is really only useful for Intel Hex files (not for Universal or UF2), to detect an old V1 Intel Hex file being flashed into a V2. Although it could also be useful for UF2 files without the family ID, but realistically the micro:bit editors will not really produce these, so those might only be produced by other tools, and those users are likely to be more advanced.

mathias-arm: Sorry I missed that it was in `source/board/microbitv2/microbitv2.c`.

I was expecting to check against something like `g_board_info.target_cfg.flash_regions[0].start` which in your case is always `0`.

microbit-carlos:
@mathias-arm just to let you know that I'm looking into this again.
Currently looking at integrating and expanding the tests from https://github.com/mbrossard/DAPLink/tree/feature/vfs-tests .

PR #1093, opened 2025-02-03
dap_vendor: Buffer MSD_Write packets into 512-byte blocks.

microbit-carlos:
Currently flashing a Universal Hex in "segments" or "blocks" format via WebUSB does not work.

This PR sends to the file streamer blocks of the same size as the VFS sector size, which is what `vfs_manager` does and some stream formats expect, like Universal Hex and in the future UF2.

This could have been implemented inside `file_stream` instead, but this way it mirrors what the MSC side is doing, buffering the 64 byte packets into 512-byte sectors before passing it to `vfs_manager`, which then sends it to `file_stream`.

One constrain is that DAPLink doesn't know when the last DAP.js `ID_DAP_MSD_Write` packet is sent, so we need to flush whatever is left in the buffer on `ID_DAP_MSD_Close`. So the "close" response could contain an error writing the last buffer.

There is also a memory hit, as the buffer takes 512 bytes of RAM. We could reuse the `usb_buffer` from `vfs_manager.c` (which is passed to `usbd_msc.c`, so that's where the 64-byte MSC packets are buffer into VFS sectors), not if you prefer to keep them separated. To be fair, using WebUSB flashing + MSD drag&drop at the same time will fail anyway, so probably wouldn't need to worry much about having two different components accessing the same buffer.

### DIFF
--- a/source/daplink/cmsis-dap/DAP_vendor.c
+++ b/source/daplink/cmsis-dap/DAP_vendor.c
@@ -38,11 +38,19 @@
 #include "settings.h"
 #include "target_family.h"
 #include "flash_manager.h"
+#include "util.h"
 #include <string.h>
 #include "daplink_vendor_commands.h"
 
 #ifdef DRAG_N_DROP_SUPPORT
 #include "file_stream.h"
+
+// Reusing the MSC sector buffer from vfs_manager.c to save memory
+// as using both at the same time will break anyway
+extern uint32_t usb_buffer[VFS_SECTOR_SIZE / sizeof(uint32_t)];
+static uint8_t *file_stream_buffer = (uint8_t *)usb_buffer;
+static const uint32_t file_stream_buffer_size = sizeof(usb_buffer);
+static uint16_t file_stream_buffer_pos = 0;
 #endif
 
 //**************************************************************************************************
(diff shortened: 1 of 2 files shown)


source/daplink/cmsis-dap/DAP_vendor.c
```diff
@@ -38,11 +38,17 @@
...
 #include <string.h>
 #include "daplink_vendor_commands.h"
 
 #ifdef DRAG_N_DROP_SUPPORT
 #include "file_stream.h"
+#include "virtual_fs.h"
+
+// TODO: It might be more memory efficient to re-use the usb_buffer from vfs_manager.c
+static uint8_t __ALIGNED(4) file_stream_buffer[VFS_SECTOR_SIZE];
+static uint16_t file_stream_buffer_pos = 0;
```
microbit-carlos, line 51: As described in the PR, this is the buffer that could be removed and `usb_buffer` from `vfs_manager.c` used instead.
Left it as a TODO comment to come to an agreement before merging the PR.

microbit-carlos:
Out of order PR + UF2 + creating a 512 byte buffer here was causing the builds fail for some ports due to inssuficient `.bss` so I had to reuse the `usb_buffer` from `vfs_manager.c` in the end.
Force pushed, diff can be seen in:
- https://github.com/ARMmbed/DAPLink/compare/15d3016c1a51e8c97785f3048338a6abb225ad88..eba229a887eb327197ea70af1c6b12297fca3b96

PR #1104, opened 2025-06-23
Support alternate SWD pins via vendor command

kbanlk:
Hi there, we have found it useful to use multiple sets of SWD pins with our ATSAM3U2C-based debug probes in our HITL setup, where we have multiple targets wired up to the same HIC to save on system cost. This feature allows using a vendor command to configure which SWD pins are used by the HIC. We can't talk to multiple targets at the same time (SWD pin selection is mutually exclusive) but this at least allows talking to one at a time without requiring an external multiplexer or changing the copper.

This could probably also be ported to other HICs; we implemented support only on ATSAM3U2C.

### DIFF
--- a/source/daplink/cmsis-dap/DAP_vendor.c.orig
+++ b/source/daplink/cmsis-dap/DAP_vendor.c.orig
@@ -41,7 +41,6 @@
 #include "util.h"
 #include <string.h>
 #include "daplink_vendor_commands.h"
-#include "daplink_swd_dynamic_pins.h"
 
 #ifdef DRAG_N_DROP_SUPPORT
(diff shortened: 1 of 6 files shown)


mbrossard:
This change is fairly invasive to the `sam3u2c` HIC and would impact other probes.  Look at how the `nrf52820` has `NRF528XX_DYNAMIC_PIN` (and the `source/board/microbitv2` that uses it) define that allows it to use global variables instead of preprocessor constants to use pins.
I would anticipate the whole feature to be gated by a define from the board, the number of pin configuration and the pin-mappings should be set by the board.

PR #1106, opened 2025-07-20
hic: DAP_PACKET_SIZE must be 64 for FS USB HICs (fixes #1105)

### LINKED ISSUE #1105: Potential bug with DAP_PACKET_SIZE configuration on STM32F103CB
I'm compiling daplink firmware for STM323F103CB.
Full version with hid+bulk+etc works fine as daplink v2.
When I compiled it without HID to use it as only bulk+cdc then I observe that USB transfer hangs up.
I've found this configuration of DAP_PACKET_SIZE in source code (hic_hal/stm32/stm32f103xb):
```
/// Maximum Package Size for Command and Response data.
/// This configuration settings is used to optimize the communication performance with the
/// debugger and depends on the USB peripheral. Typical vales are 64 for Full-speed USB HID or WinUSB,
/// 1024 for High-speed USB HID and 512 for High-speed USB WinUSB.
#ifndef HID_ENDPOINT            //HID end points currently set limits to 64
#define DAP_PACKET_SIZE         512              ///< Specifies Packet Size in bytes.
#else
#define DAP_PACKET_SIZE         64              ///< Specifies Packet Size in bytes.
#endif
```
STM32F103CB has Full-speed USB so USB packet size should be set to 64B not 512B.
I changed it to 64 and now compiled firmware works.
I think no one found this bug earlier because a full build uses HID, and then the packet size is fortunately set to 64.

mbrossard:
All HICs that only support FS USB should only have 64 bytes as DAP_PACKET_SIZE value.

### DIFF
--- a/source/hic_hal/freescale/kl26z/DAP_config.h
+++ b/source/hic_hal/freescale/kl26z/DAP_config.h
@@ -84,11 +84,7 @@ This information includes:
 /// This configuration settings is used to optimize the communication performance with the
 /// debugger and depends on the USB peripheral. Typical vales are 64 for Full-speed USB HID or WinUSB,
 /// 1024 for High-speed USB HID and 512 for High-speed USB WinUSB.
-#ifndef HID_ENDPOINT            //HID end points currently set limits to 64
-#define DAP_PACKET_SIZE         512              ///< Specifies Packet Size in bytes.
-#else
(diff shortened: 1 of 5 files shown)


microbit-carlos (approved):
LGTM 👍
Also the nRF52820 & nRF52833 are Full Speed USB 2.0.

PR #1113, opened 2026-07-09
MAX32625: USBD Stack Improved for better USB-Serial Communication

ahmetalincak:
- Do not reset USB baudrate after a bus reset event received to preserve existing UART comm.
- Remove circ_buf calls to avoid disabling interrupts too frequently
- Modify USB, UART and Timer interrupt priority
- Update SOF Event handling mechanism


ahmetalincak:
@mathias-arm Could you please review and merge this if it's okay?

ahmetalincak:
Hey @mathias-arm. This is an internally reviewed PR within ADI.

PR #1114, opened 2026-08-17
[HIC]Add support for the NXP LPC845 Breakout board

HuanQian571:
Adds DAPLink interface firmware support for the **NXP LPC845 Breakout board** (HIC: LPC11U35, target: LPC845), plus two fixes found while bringing it up.

### Commits

**1. `lpc11u35: Fill the interface image out to the end of the app region.`**

The LPC11U35 has no DAPLink bootloader, so the only way to program the HIC is the on-chip ROM ISP MSC drive (`CRP DISABLD`). That drive maps a fixed-size `firmware.bin` 1:1 onto the whole `0xF000` app region, and deleting `firmware.bin` only clears the directory entry — the flash sectors keep their previous contents. A short image therefore left stale bytes from the old firmware in the tail, producing an image that would not boot, with no USB enumeration and no indication why.

Adds a `.fill` section padding `.text` out to the end of the app region, the same way `kinetis.ld` and `lpc4322.ld` already do.

This also fixes the interface CRC reported in `DETAILS.TXT`. `info_crc_compute()` checksums `DAPLINK_ROM_IF_SIZE - 4` unconditionally, so with a short image the value depended on whatever stale bytes followed the image and never matched the CRC recorded at build time. The build-time CRC now lands at `0xEFFC`, exactly where `info_crc_compute()` expects it.

Affects all 30 `lpc11u35_*_if` projects, so it is kept as a separate commit.

**2. `Add support for the NXP LPC845 Breakout board.`**

Board and target-family layers only; the firmware reuses the existing `lpc11u35` HIC anchor. No custom reset or unlock behaviour is needed, so the generic `kStub_HWReset_FamilyID` stub family is used rather than a new `target_reset.c`.

**3. `lpc845: Regenerate the flash algorithm from the official LPC84x FLM.`**

The flash algorithm was initially copied from the LPC824 family on the assumption that LPC84x shares the LPC82x IAP ROM interface. It does not. Keil's own algorithm source, shipped in `Keil.LPC800_DFP` as `Flash/LPC8xx_IAP/FlashPrg.c`, distinguishes the two:

```c
#if   defined LPC8XX || defined LPC8N04
  #define IAP_Call ((IAP_Entry) 0x1FFF1FF1)
#elif defined LPC84X || defined LPC80X
  #define IAP_Call ((IAP_Entry) 0x0F001FF1)
```

Byte-comparing `LPC8xx_32.FLM` against `LPC84x_64.FLM` shows exactly three functional differences, all of which the copied blob got wrong:

| | LPC82x (was used) | LPC84x (correct) | effect |
|---|---|---|---|
| IAP entry | `0x1FFF1FF1` | `0x0F001FF1` | LPC84x boot ROM is at `0x0F000000`; every IAP call branched to a non-existent address |
| `END_SECTOR` | `31` | `63` | chip erase covered only the lower 32 kB, leaving the upper 32 kB unerased |
| `MAINCLKSEL`/`MAINCLKUEN`/`SYSAHBCLKDIV` | SYSCON `+0x70/74/78` | `+0x50/54/58` | `Init` wrote to the wrong registers |

The `END_SECTOR` bug is on the default drag-n-drop path: `flash_manager` calls `erase_chip()` whenever `kEnablePageErase` is not set, and this board does not set it.

Regenerated from the authoritative algorithm rather than patching bytes:

```
python tools/generate_flash_algo.py LPC84x_64.FLM \
    --blob-start 0x10000000 --pack-path Keil.LPC800_DFP.1.10.2.pack
```

The flash geometry the previous blob declared was already correct and is unchanged (one 1 kB sector descriptor at 0, 64 kB region), matching the FLM. `target.c` needs no change; its flash and RAM regions also match the pack's `IROM1`/`IRAM1` for `LPC845M301JBD48`.

### Testing

- All 30 `lpc11u35_*_if` projects build with GCC 10.3-2021.10; each produces a `0xF000` image and none overflow `m_text`.
- Flashed `lpc11u35_lpc845breakout_if` to an LPC845 Breakout board. It enumerates as a full DAPLink MSD drive, and `DETAILS.TXT` reports an interface CRC matching the build's `_crc.txt`.
- Drag-n-drop to the LPC845 target succeeds. Reading the target flash back over SWD shows a ~34 kB image programmed correctly past `0x8000` — sectors that the previous `END_SECTOR` of 31 would have left unerased.
- CMSIS-DAP/SWD verified via pyOCD against `lpc845m301jbd48`: halt/resume, core register reads, CPUID (`0x410CC601`, Cortex-M0+), 4 kB RAM write/read round trip, and `reset_and_halt`.

`test/run_test.py` endpoint tests were not run: they require a `daplink-validation` image built for the LPC845, which does not exist, and `--testdl` self-skips because this board has no bootloader.

### Notes for reviewers

- A `test/info.py` entry is added for board ID `0xD002` (`LPC845BREAKOUT`).
- `records/board/lpc845breakout.yaml` and `source/board/lpc845breakout.c` follow `docs/PORT_BOARD.md`; no common code is touched, so the no-`#ifdef` rule is preserved.

### DIFF
--- a/source/board/lpc845breakout.c
+++ b/source/board/lpc845breakout.c
@@ -1,0 +1,32 @@
+/**
+ * @file    lpc845breakout.c
+ * @brief   board ID for the NXP LPC845 Breakout board
+ *
+ * DAPLink Interface Firmware
+ * Copyright (c) 2009-2019, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "target_family.h"
+#include "target_board.h"
+
+const board_info_t g_board_info = {
+    .info_version = kBoardInfoVersion,
+    .board_id = "D002",
+    .family_id = kStub_HWReset_FamilyID,
+    .target_cfg = &target_device,
+    .board_vendor = "NXP",
+    .board_name = "LPC845BREAKOUT",
+};
--- a/test/info.py
+++ b/test/info.py
@@ -245,6 +245,7 @@ SUPPORTED_CONFIGURATIONS = [
     (   0x9905,     VENDOR_TO_FAMILY('Nordic', 2),      'nrf52820_microbit_if',                     'nrf52820_bl',      'Microbitv2'                            ),
     (   0x9906,     VENDOR_TO_FAMILY('Nordic', 2),      'nrf52820_microbit_if',                     'nrf52820_bl',      'Microbitv2'                            ),
     (   0xA127,     VENDOR_TO_FAMILY('Ambiq', 1),       'kl26z_artemis_dk_if',                      'kl26z_bl',         'ARTMBED'                               ),
+    (   0xD002,     VENDOR_TO_FAMILY('Stub', 1),        'lpc11u35_lpc845breakout_if',               None,               'LPC845BREAKOUT'                        ),
     (   0x0000,     VENDOR_TO_FAMILY('Stub', 1),        'kl26z_if',                                 None,               None                                    ),
     (   0x0000,     VENDOR_TO_FAMILY('Stub', 1),        'k20dx_if',                                 None,               None                                    ),
     (   0x0000,     VENDOR_TO_FAMILY('Stub', 1),        'k26f_if',                                  None,               None                                    ),
--- a/projects.yaml
+++ b/projects.yaml
@@ -410,6 +410,10 @@ projects:
         - *module_if
         - *module_hic_lpc11u35
         - records/board/lpc824xpresso.yaml
+    lpc11u35_lpc845breakout_if:
+        - *module_if
+        - *module_hic_lpc11u35
+        - records/board/lpc845breakout.yaml
     lpc11u35_mini_iot_lpc54018_if:
         - *module_if
         - *module_hic_lpc11u35
--- a/records/board/lpc845breakout.yaml
+++ b/records/board/lpc845breakout.yaml
@@ -1,0 +1,6 @@
+common:
+    sources:
+        board:
+            - source/board/lpc845breakout.c
+        family:
+            - source/family/nxp/lpc845/target.c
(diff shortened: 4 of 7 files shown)


source/board/lpc845breakout.c
```diff
@@ -0,0 +1,32 @@
...
+#include "target_family.h"
+#include "target_board.h"
+
+const board_info_t g_board_info = {
+    .info_version = kBoardInfoVersion,
+    .board_id = "0845",
+    .family_id = kStub_HWReset_FamilyID,
+    .target_cfg = &target_device,
+    .board_vendor = "NXP",
+    .board_name = "LPC845 Breakout",
```
mathias-arm: Should this be `LPC845BREAKOUT` (based on https://www.keil.arm.com/boards/nxp-lpc845breakout-bc00216/projects/)?

source/board/lpc845breakout.c
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "target_family.h"
+#include "target_board.h"
+
+const board_info_t g_board_info = {
+    .info_version = kBoardInfoVersion,
+    .board_id = "0845",
```
mathias-arm: It seems 0x0845 is already pre-allocated: https://github.com/pyocd/pyOCD/blob/main/pyocd/board/board_ids.py#L197

The process for requesting Mbed Board IDs has now disappeared. I could allocate you one in 0xD0000 (DAPLink-space): https://github.com/ARMmbed/DAPLink/wiki/DAPLink-board-IDs

test/info.py
```diff
@@ -194,6 +194,7 @@ def VENDOR_TO_FAMILY(x, y) : return (VENDOR_ID[x] <<8) | y
     (   0x0816,     VENDOR_TO_FAMILY('Stub', 1),        'stm32f103xb_stm32f746zg_if',               'stm32f103xb_bl',   'ST-Nucleo-F746ZG'                      ),
     (   0x0824,     VENDOR_TO_FAMILY('Stub', 1),        'lpc11u35_lpc824xpresso_if',                None,               'LPCXpresso824-MAX'                     ),
     (   0x0835,     VENDOR_TO_FAMILY('Stub', 1),        'stm32f103xb_stm32f207zg_if',               'stm32f103xb_bl',   'ST-Nucleo-F207ZG'                      ),
+    (   0x0845,     VENDOR_TO_FAMILY('Stub', 1),        'lpc11u35_lpc845breakout_if',               None,               'LPC845-BRK'                            ),
```
mathias-arm: This also needs to be changed.

mathias-arm (changes requested):

source/board/lpc845breakout.c
HuanQian571: Thanks for the review. Yes, it's better to align the board name with official one.  updated.

source/board/lpc845breakout.c
HuanQian571: I see 0xD000 is also allocated by NM180410 https://github.com/pyocd/pyOCD/blob/main/pyocd/board/board_ids.py#L320 ,  can I use '0xD001' for lpc845-brk?

mathias-arm: I allocate 0xD002 to this board.

HuanQian571: updated board id to 0xD002

mathias-arm (approved):

